PR #4, opened 2017-05-13
Remove whitespace from RSS links

erengy:
Might cause issues for applications with an XML parser that considers leading and trailing whitespace characters as significant. Encountered and fixed this issue in my application, sending this PR in case it affects others.

### DIFF
--- a/nyaa/templates/rss.xml
+++ b/nyaa/templates/rss.xml
@@ -10,9 +10,7 @@
 		{% if torrent.has_torrent %}
 		<item>
 			<title>{{ torrent.display_name }}</title>
-			<link>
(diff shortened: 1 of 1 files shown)


TheAMM:
If you'll also do the same for the description (for neatness), I'll merge this! (Or someone else if I'm sleeping)
Good find, though, lots of things were just *done* in a hurry.

erengy:
Done. Note that you might want to update the description, because it's displayed as `RSS Feed for /`. The underlying issue seems to be [this line](https://github.com/nyaadevs/nyaa/blob/f7d14ee19f750fe819f92905edafc024633b9780/nyaa/routes.py#L255), where you're passing `/` as the label/term.

PR #14, opened 2017-05-13
Improved tracker list

wranai:
I looked through the trackers on newtrackon.com. These are trackers with great uptime (mostly better than 99%), but also preferring UDP and IPv6 (because why not).

Pre-existence in the magnet links from different Nyaa mirrors was important, but I tried to get rid of the sub-standard ones, or those that simply didn't work.


wranai:
That would be a pity, it seems like a very good tracker... Also, do normal people actually look through the tracker list? Those who did are those who would already know, or would know how to check. But if you still think so, free to remove it.

aldacron:
Explodie is blacklisted from Tokyotosho

PR #15, opened 2017-05-13
Replace hardcoded URLs

wranai:


nyaadev:
I think these are the only actually needed hardcoded urls.
Since one page redirects to the other.

wranai:
lol so true sry

PR #16, opened 2017-05-13
Search for the URL bar and Google search box

wranai:


sharkykh:
@wranai 
On Chrome (after I remove Nyaa.si from the search engines list),
as soon as I search once on the site, it automatically adds the search engine again.

So what does this do?

wranai:
Yea, I noticed that just recently; it must be something with how the search form is named, because Chrome doesn't pick up on e.g. Archive.org's search box...

It would (could) also make Google add a search box if nyaa.si is the first hit for a search, but idk if that's important.

Just delete it if you think it's not important.

katnyaa:
The site's `robots.txt` prevents crawling anyway.

PR #17, opened 2017-05-13
Small CSS tweaks to the torrent listing and torrent view on mobile

Ricardo1991:
Made it so that the torrent info is a bit easier to read on the compact version.

- [Mobile before](http://i.imgur.com/L6TyR3h.png)    
- [Mobile after](http://i.imgur.com/euutdTJ.png)

Also removed the slightly gray background on odd numbered torrents on the list.


nyaadev:
If possible the information field should be above the seeders.

Ricardo1991:
That is not something that should be fixed via CSS, considering the rows on that table don't even have css classes. 

http://i.imgur.com/H37LE3j.png

PR #22, opened 2017-05-13
Make upload form input label less ambiguous.

decimoe:
Change "Display name" label in torrent upload form to "Torrent name", so as not to confuse the minority of users who think their username goes there.


nyaadev:
oops I already fixed this

Issue #24, opened 2017-05-13
Make Information link clickable


marconii2002:
Can you make information website or irc link clicakable like in old Nyaa ?

HitmanJ:
+1

TheAMM:
Done.

PR #30, opened 2017-05-13
Changed information link to hyperlink (#24)

adsko:
I changed information link to hyper link.

More information in: issue #24


nyaadev:
Not every information field is a link though, having a simple regex check if it is and then doing what you did would be a solution.

TheAMM:
Implemented myself with IRC channel formatting included, see 00a0327b192ca931b0ee018112fb9b345396f3e1. Thanks for getting involved, anyway!

Issue #26, opened 2017-05-13
Filter is not saved after page reloading


UselessToucan:
```
STEPS TO REPRODUCE:
1. Open https://nyaa.si
2. Perform search request with category and filter. (for example: anime, trusted only)
3. Check <select> elements state

EXPECTED RESULT:
Category and filter values are preserved after page reloading.

ACTUAL RESULT:
Filter value is not preserved. After page reloading it is always set to "Show all".
```

TheAMM:
Fixed in 8d244675a0ec6cb60bf70dfc0cd5d15260520a51.

PR #33, opened 2017-05-13
Fixed problem with saving filter after reloading (#26)

adsko:
There was a simple problem with dictionary field name.


TheAMM:
Ah damn, that was my fault in 43406fb831596fb363f600b80c3f6de61a4c5934, missed those in a search.

PR #46, opened 2017-05-14
Markdown preview

nathancyam:
In order to make the markdown editor fields reusable, I've added a `render_markdown_editor` macro that should automatically generate the relevant DOM structure which is utilised by the changes in `main.js`. To make sure that we can multiple markdown editors, you have to namespace them with the field's name.

Let me know you would like any further changes.


TheAMM:
Remember `edit.html` as well! I'll merge after that's done too..
Otherwise looks good; could do with borders around the preview, but that's not a big problem. Congrats on what's probably the first grand feature-through-PR!
(However, I may still go for the JS-only [well, plus one attribute to the text field] thingy I visioned, the idea just feels neat)

TheAMM:
Yes, very nice!

PR #50, opened 2017-05-14
elasticsearch-based search (WIP)

qqueue:
explanation soon™


uro:
The way where this project is going is very messy, this what you are doing is just an implementation of search functionality. For me it should not made any changes in frontend parts. Maybe we should use some DI container?

aldacron:
@uro I agree it's very messy, but we're trying to iterate quickly. Right now the search isn't very accurate and/or fast, so we want to migrate to using a real FTS solution.

PR #51, opened 2017-05-14
Update RSS template

### LINKED ISSUE #34: rss feed not working with sonarr
I get the error message in Sonarr, 'Indexer feed is not supported: Feed doesn't contain the release content size.'. Is it possible to add the content size to make it work, please.

sharkykh:
This is intended to fix #34, and also for other auto-downloaders (SickBeard/SickRage/Medusa/etc.)
(**Supersedes #31**)

* Refactor:
  * Only check `if torrent.has_torrent` for the `<link>` element.
* Add elements:
  * description
  * category (formatted as `main_category - sub_category`)
  * size (formatted size) ~_**(in bytes)**_ **I wasn't sure what's better** - size in bytes or formatted size?~
  * seeders
  * leechers
  * downloads
* Add bare torrent hash:
  I'm guessing sometimes users need the torrent hash
  (magnet_uri only shows up when no torrent is available)
  lambda is probably not the best way to go,
  again, wasn't sure - to add a function just for this, to add it in the torrent model, to add a function to `torrents.py`?

### Any suggestions for improvement are welcome!

~I've seen several RSS feeds that use `CDATA` in order to avoid parsing of tags (HTML for example) within the XML document.
I didn't add this, but it should be considered - for data entered by users (`torrent.display_name` and `torrent.description`) and for `magnet_uri`~
```xml
<![CDATA[your text here]]>
```

### DIFF
--- a/nyaa/routes.py
+++ b/nyaa/routes.py
@@ -303,6 +303,7 @@ def render_rss(label, query, use_elastic):
                                     use_elastic=use_elastic,
                                     term=label,
                                     site_url=flask.request.url_root,
+                                    compute_hash=lambda x: base64.b32encode(x).decode('utf-8'),
                                     torrent_query=query)
     response = flask.make_response(rss_xml)
     response.headers['Content-Type'] = 'application/xml'
--- a/nyaa/templates/rss.xml
+++ b/nyaa/templates/rss.xml
@@ -5,33 +5,31 @@
 		<link>{{ url_for('home', _external=True) }}</link>
 		<atom:link href="{{ url_for('home', page='rss', _external=True) }}" rel="self" type="application/rss+xml" />
 		{% for torrent in torrent_query %}
-		{% if torrent.has_torrent %}
 		<item>
 			<title>{{ torrent.display_name }}</title>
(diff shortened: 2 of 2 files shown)


snowfag:
Magnets and torrents can't be mixed together in the same rss feed because not everything can handle both provided it's only setup to handle one (e.g. flexget downloading into a watch folder instead of using something like the transmission rpc (which is pretty unreliable))

Nazonaze:
Descriptions could probably be left off completely since the RSS is primarily intended for torrent clients where it shouldn't even matter.

But if included it should definitely be wrapped in CDATA.

sharkykh:
@Nazonaze 
> Descriptions could probably be left off completely since the RSS is primarily intended for torrent clients where it shouldn't even matter.

That's a good point, thanks.
> But if included it should definitely be wrapped in CDATA.

I'll wrap in CDATA anyway.

If you decide you don't want descriptions, then I'll remove the element.

sharkykh:
@TheAMM 
Can you review this too, please?

sharkykh:
Rebased onto 00c768c722117ac1edf99b44263785230f5b0e17.

PR #55, opened 2017-05-14
Remove some trackers because they're down

Kcchouette:
- exodus.desync.com
- tracker.sktorrent.net

Are down on my side.

Thanks


sharkykh:
@Kcchouette This is a dupe since they're both removed in #14,
no need for two pull requests.

aldacron:
Closed by merging #14

Issue #56, opened 2017-05-14
Dark Mode bootstrap 404


Arsobbiak:
After reload page in Dark theme its try to download:

`https://nyaa.si/static/css/bootstrap.min.css.map`

but this is not exist.

EduenSarceno:
I confirm this, when dark mode is on, the list of http requests is:


![captura de pantalla de 2017-05-14 07-55-37](https://cloud.githubusercontent.com/assets/18363635/26034510/f67afe12-387a-11e7-9477-61ccde65133d.png)

![captura de pantalla de 2017-05-14 07-53-19](https://cloud.githubusercontent.com/assets/18363635/26034486/75cbd476-387a-11e7-93e3-23c60c4d4d61.png)

aldacron:
Fixed

PR #58, opened 2017-05-14
Remove sourceMappingURL from bootstrap-dark.min.css

JSFernandes:
This PR fixes the issue #56 

Some browsers, such as Firefox 53.0 try to download the specified file in sourceMappingURL, to improve development experience. Since we don't host this file, some users were making a request resulting in 404 when loading the page in dark mode.


uro:
Maybe better way is to fix this dark mode? Could be very useful. @JSFernandes

JSFernandes:
@uro can you elaborate? Dark mode is working as intended.

We could add `bootstrap.min.css.map` to fix the 404, which would allow us to see/edit un-minified Bootstrap CSS, but we aren't doing it for the other Boostrap CSS files, so why do it only for `bootstrap-dark`?

uro:
Ok, I understand your PR right now. But it's still better to use gulp or something to prepare assets for frontend usage, maybe in the feature.
Thanks

PR #67, opened 2017-05-15
Elasticsearch

qqueue:
recreating as a PR for tracking.


TheAMM:
Closed for branch shenanigans.

TheAMM:
No hold up

ayyaruq:
There are some comments in the PR about reindexing and performance and whatnot. You'd probably want to split your index by an easy modifier, like language or date. You get much better search performance that way, and you can reindex a given day of data or rebuild an individual index in the background without breaking search entirely while it happens. Something like `nyaa-%YYYY-%MM-%DD` would be fine, since you can search in ES with a wildcard for the index.

As far as dereferencing goes, you can make a nested index for user data and just update the child when a user changes their name but the ID is the same. Searching a nested index is about as difficult as a normal primary. It might also be worth setting analysers explicitly so you can do things like text suggestion for group or user names, and adding specific fields such as seed/peer counts to not be tokenised or indexed to prevent massive update latency. Non-indexed fields can still be used for sorting and ordering, within limits of pagination.

For importing from the DB, the bulk API might help. Doing torrents one at a time is a bit silly when ES can do 10k updates and merge them to the log as one write. The `sync_to_es` script is I think unnecessary too. Just have a hook on DB updates (or an event capture on upload) to index the new document when appropriate. It will only have as much latency in searching as it takes for it to get indexed and replicated to quorum.

I would probably also disable replication (update index settings) before bulk importing since it will make the process much faster. If you have a split index, it's even easier. A cron job to set replication higher for the last 7 days of indices could also make searching more recent content more performant.

import_to_es.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python
+"""
+Bulk load torents from mysql into elasticsearch `nyaav2` index,
+which is assumed to already exist.
```
ayyaruq: This doesn't really matter. Unless you have a custom mapping, ES will create a new index on the fly when you `PUT` or `POST` to it.

qqueue:
Ah I see you are an elasticsearch user. I am pretty elastic myself.

>you'd probably want to split your index by an easy modifier, like language or date.

Soon™. currently deployment is on a single relatively beefy machine, and there are only like 800k torrents total, so it didn't seem worth it to do more than one index and one shard with no replication. The full bulk import only takes ~2 minutes on my machine. Also, the update rate is going to be fairly minimal, on the scale of 1 new import per minute, so I'll think the next thing to scale out would be read replicas. logstash indices this ain't.

>As far as dereferencing goes, you can make a nested index for user data and just update the child when a user changes their name but the ID is the same

You mean parent/child relationship, right? Have you actually done user/post-style modeling with parent/child? It seems like it should work, but I've never done it, so advice and examples there would be helpful. I feel like there are weird caveats or gotchas with parent/child, but never had to run into them.

Currently search is only going to support the display_name field, so indexing that text user names is pointless. Would be useful later though, also if we do stuff like facet aggregation. Depends on @aldacron's product schedule.

>it might also be worth setting analysers explicitly

oh yeah, @aldacron you need to commit your explicit mapping.yml.

>Non-indexed fields can still be used for sorting and ordering, within limits of pagination.

right, should do that for seed/leecher counts, since we don't need the reverse index.

>For importing from the DB, the bulk API might help. Doing torrents one at a time is a bit silly when ES can do 10k updates and merge them to the log as one write.

That's what the import to es script is doing.

> The sync_es script is I think unnecessary too. Just have a hook on DB updates (or an event capture on upload) to index the new document when appropriate.

the sync_es script _is_ getting hooks on db updates via binlog. I don't think it needs to be batched because again new/edited torrents will be relatively infrequent (1 per minute). The seeder/leecher updates _may_ need to be batched depending on how often those get touched in mysql by the tracker. If we end up doing a /scrape poll every N minutes though, that should also probably be fine. I guess the binlog WriteRowsEvent does give you multiple rows which could be batch-sent to es.

>I would probably also disable replication (update index settings) before bulk importing since it will make the process much faster. 

Yep, though I don't think we're going to use replication in prod because cheaper. Would be good to tun off index refresh during that import script though, will add that. Note that the plan is to only ever have to run the initial import once (modulo failures or mapping changes requiring reindexing), with the binlog-tailing sync script running continuously in the meantime. If the binlog script dies or es dies temporarily, than you should be able to pick up the same position in the binlog. quality enterprise archecture, yo.

ayyaruq:
Oh you merged this haha. Binlog is a bit different to hooking. I meant inside SQAlchemy, like how in ActiveRecord or something you could have a callback registered to `:on_save` events. I mustn't have read the import script properly though, Python isn't my main language of choice.

Having smaller indices doesn't just give you fast rebuilds or huge amounts of data flexibility. Since you can't update a mapping without recreating the index, it gives you an easier way to make index changes. You can also use an index alias, like `torrents => torrents_v3` and create a new index on demand for more static indices, then just point the alias to the new version after getting data into it.

I still recommend sharding even without replication, it makes writes faster. Replication on the other hand makes reads faster so is probably a bigger priority, but you can set that whenever you want.

The last time I did nested indices was for embedding relevant information inside the parent object, in my case a user object that had purchase history embedded. For post comments, I would probably have a thread ID and keep them as separate documents in a vaguely linked list. Torrent "fixed" data such as the name, uploader, or hash as the primary document with "flexible" data such as seed count or reliability of a given tracker as the child would make sense though.

PR #74, opened 2017-05-16
Add full torrent name to link "title"

Ricardo1991:
To solve a problem pointed out by someone in the IRC.
User could not read the full name of the torrent

Image

So this will allow the full name to appear on mouseover.

----

**I'll just ask someone to test, since I don't actually have the project and database set up.** For example, does it need the `| escape`?


TheAMM:
There's something funny going on with your commits - you seem to be using `git pull` to sync up the repo when you have local commits - you should really be doing the following to fetch changes and apply them without filling the history with useless merges:
```
git fetch upstream
git rebase upstream/master
```
(where upstream has been added with `git remote add git@github.com:nyaadevs/nyaa.git`)

The title text is a good point, but please see if you can get rid of the seven other commits. If you need help, ask. (If you don't feel like it, I can just create a commit of this myself, but there is value in you learning the tricks of git)
And yes, add in the `| escape`. The Jinja2 environment actually should autoescape it, but explicit is better than implicit.

PR #81, opened 2017-05-16
Add full torrent name to link "title"

Ricardo1991:
Removed my old PR #74 because it had become a mess.

---
To solve a problem pointed out by someone in the IRC.
User could not read the full name of the torrent

Image

So this will allow the full name to appear on mouseover.

----

**I'll just ask someone to test, since I don't actually have the project and database set up.**

### DIFF
--- a/nyaa/templates/search_results.html
+++ b/nyaa/templates/search_results.html
@@ -60,9 +60,9 @@
 				</a>
 				</td>
 				{% if use_elastic %}
-                <td><a href="{{ url_for('view_torrent', torrent_id=torrent.meta.id) }}">{%if "highlight" in torrent.meta %}{{ torrent.meta.highlight.display_name[0] | safe }}{% else %}{{torrent.display_name}}{%endif%}</a></td>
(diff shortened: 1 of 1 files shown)


nyaa/templates/search_results.html
```diff
@@ -60,9 +60,9 @@
 				</a>
 				</td>
 				{% if use_elastic %}
-                <td><a href="{{ url_for('view_torrent', torrent_id=torrent.meta.id) }}">{%if "highlight" in torrent.meta %}{{ torrent.meta.highlight.display_name[0] | safe }}{% else %}{{torrent.display_name}}{%endif%}</a></td>
+                <td><a href="{{ url_for('view_torrent', torrent_id=torrent.meta.id) }}" title="{{ torrent.display_name | escape }}">{%if "highlight" in torrent.meta %}{{ torrent.meta.highlight.display_name[0] | safe }}{% else %}{{torrent.display_name}}{%endif%}</a></td>
```
sharkykh: I know it's not your doing, but please replace these leading spaces with tabs (like all the other lines)

PR #93, opened 2017-05-17
Put tracker URL in <kbd>

sharkykh:
Looks nicer (IMO)
***
image


FichteFoll:
It'd be better if this was a span with class styling rather than abusing html5 tags.

sharkykh:
Why abusing?
`kbd` is keyboard input. That's what it is, isn't it?

FichteFoll:
Never mind, I was confusing tags.

PR #94, opened 2017-05-17
Fix torrent view text overflowing

Ricardo1991:
Fixes this issue  pointed on on the IRC

### DIFF
--- a/nyaa/static/css/main.css
+++ b/nyaa/static/css/main.css
@@ -99,6 +99,12 @@ table.torrent-list thead th.sorting_desc:after {
 	}
 }
 
+@media (min-width: 992px){}
+	.col-md-5 {
+		word-wrap: break-word;
+	}
+}
+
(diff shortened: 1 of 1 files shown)


nyaa/static/css/main.css
```diff
@@ -94,17 +94,23 @@ table.torrent-list thead th.sorting_desc:after {
...
 	.col-md-5 {
-	    margin-left: 20px;
-	    margin-bottom: 10px;
+		margin-left: 20px;
+		margin-bottom: 10px;
+	}
+}
+
+@media (min-width: 992px){
+	.col-md-5 {
```
sharkykh: I don't think `.col-md-5` is a good selector for this.
Maybe `.panel-body .col-md-5`?

Ricardo1991: Yup, you are right. Fixed

PR #95, opened 2017-05-17
Auto python lint check using Travis

sharkykh:
...So that no one will forget to adhere to PEP8 coding style.

Of course repo owners need to enable the [Travis-CI](https://github.com/integrations/travis-ci) integration for this organization/repo for it to test every commit and pull request.

Output example (of a failed lint check):
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: python
+
+python:
+  - 3.6
+
+dist: xenial
+sudo: false
+
+cache: pip
+
+install:
+  - pip install --upgrade pep8
+
+script:
+  - pep8 nyaa/ --show-source --max-line-length=100
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,3 +13,6 @@ install:
 
 script:
   - pep8 nyaa/ --show-source --max-line-length=100
+
+notifications:
+  email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,10 @@ sudo: false
 cache: pip
 
 install:
-  - pip install --upgrade pep8
+  - pip install --upgrade pycodestyle
 
 script:
-  - pep8 nyaa/ --show-source --max-line-length=100
+  - pycodestyle nyaa/ --show-source --max-line-length=100
 
 notifications:
   email: false


FichteFoll:
The pep8 package has been renamed to pycodestyle  while ago.

sharkykh:
@FichteFoll Thanks, I'll change it.

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+dist: xenial
+sudo: false
+
+cache: pip
+
+install:
+  - pip install --upgrade pycodestyle
+
+script:
+  - pycodestyle nyaa/ --show-source --max-line-length=100
```
mc10: I think you should call `./lint.sh` and fix the script call in that file, since the file already exists.

.travis.yml
```diff
@@ -0,0 +1,18 @@
+language: python
+
+python:
+  - 3.6
+
+dist: xenial
```
mc10: Curious: is there a reason why `xenial` is specifically specified here instead of `trusty` or just left blank?

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+sudo: false
+
+cache: pip
+
+install:
+  - pip install --upgrade pycodestyle
```
mc10: Probably overkill for now, but we could just use `pip install -r requirements.txt`, which will also set up Travis for any future testing.

.travis.yml
sharkykh: Lint script runs `autopep8` 
I didn't like the report style provided by it.
Also it is based on pycodestyle with some minor additions AFAIK.

.travis.yml
sharkykh: Before opening the PR I used `trusty` explicitly. I use `xenial` on another repo and it felt like the container was a bit faster to initialize.

.travis.yml
sharkykh: Way overkill, but yes.
Again, for the sake of speed, we shoulsn't just now, since there are no tests as of yet.

shm0o: Travis runs `pip install -r requirements` by default on python projects, so you should maybe evaluate again what's overkill here or not.

sharkykh: @shm0o 
True, but I'm overriding this default by setting the value for the `install` key,

see https://docs.travis-ci.com/user/languages/python/#pip

> By default Travis CI uses `pip` to manage python dependencies. If you have a `requirements.txt` file, Travis CI runs `pip install -r requirements.txt` during the `install` phase of the build.
> 
> You can manually override this default `install` phase, for example:

shm0o: Ah right, I completely missed that.

PR #103, opened 2017-05-17
Move to Google Noto Sans font family and 1 more

UnKnoWn-Consortium:
1. It has Japanese fonts and is more legible than the Helvetica family.
2. Bolded navbar-brand to give it, well, a more "bold" look :)

### DIFF
--- a/nyaa/static/css/main.css
+++ b/nyaa/static/css/main.css
@@ -1,8 +1,10 @@
 @import url('//fonts.googleapis.com/earlyaccess/notosansjapanese.css');
+@import url('//fonts.googleapis.com/earlyaccess/notosanstc.css');
+@import url('//fonts.googleapis.com/earlyaccess/notosanssc.css');
 @import url('//fonts.googleapis.com/css?family=Noto+Sans');
-
+  
 body {
-    font-family: 'Noto Sans', 'Noto Sans Japanese', sans-serif;
+    font-family: 'Noto Sans', 'Noto Sans Japanese', 'Noto Sans TC', 'Noto Sans SC', sans-serif;
 }
 
 nav a.navbar-brand {
--- a/nyaa/static/css/main.css
+++ b/nyaa/static/css/main.css
@@ -1,8 +1,3 @@
-@import url('//fonts.googleapis.com/earlyaccess/notosansjapanese.css');
-@import url('//fonts.googleapis.com/earlyaccess/notosanstc.css');
-@import url('//fonts.googleapis.com/earlyaccess/notosanssc.css');
-@import url('//fonts.googleapis.com/css?family=Noto+Sans');
-  
 body {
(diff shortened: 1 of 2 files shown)


nyaa/static/css/main.css
```diff
@@ -1,3 +1,14 @@
+@import url(//fonts.googleapis.com/earlyaccess/notosansjapanese.css);
+@import url('https://fonts.googleapis.com/css?family=Noto+Sans');
```
sharkykh: Since http://nyaa.si always redirects to https://nyaa.si,
I think you can drop the `https:` like on the first URL.

nyaa/static/css/main.css
```diff
@@ -1,3 +1,14 @@
...
+
+body {
+    font-family: 'Noto Sans', 'Noto Sans Japanese', sans-serif;
+}
+
+nav a.navbar-brand {
+    font-weight: bold;
```
sharkykh: Use tabs instead of spaces as per the rest of the file.

nyaa/static/css/main.css
```diff
@@ -1,3 +1,14 @@
+@import url(//fonts.googleapis.com/earlyaccess/notosansjapanese.css);
+@import url('https://fonts.googleapis.com/css?family=Noto+Sans');
+
+body {
+    font-family: 'Noto Sans', 'Noto Sans Japanese', sans-serif;
```
sharkykh: Here too.

UnKnoWn-Consortium:
@aldacron 

Before: 



After:

benwa:
As an FYI, Bootstrap is [reversing their font choices](https://v4-alpha.getbootstrap.com/content/reboot/#native-font-stack) for v4 (Coming Soon<sup>TM</sup>).  
However I'm not sure how well Japanese characters are going to look.

If optimal performance is the goal, then the least amount of external resources the better.

Nazonaze:
Noto Sans actually has the goal of "aiming to support all languages with a harmonious look and feel" so Japanese characters are very much supported.

I was actually planning on making use of it in the site design later anyway, although as part of a bigger overhaul rather than just changing the font on the existing design.

UnKnoWn-Consortium:
The Noto Sans fonts and css I used in the commit are provided by Google Fonts which is highly available and contains proper cache-control header: "public, max-age=31536000". Most if not all modern browsers would cache and reuse the fonts as instructed. Apart from that, modern browsers only request fonts that are explicitly used in the markup. In a page with only latin characters, neither the Japanese nor the Chinese fonts will be requested (and thus downloaded if unavailable from cache). With our modern web, the performance impact should have been kept to the minimum. As for character range, I have included the authentic latin font set plus its Japanese, traditional Chinese, and simplified Chinese variants which are already cherry-picked by Google to serve on its web font API.

PR #107, opened 2017-05-17
Make profile page more profile

UnKnoWn-Consortium:
1. Changed page title to "Profile of {{ USERNAME }}"
2. User information is now shown at page top (not inside one of the tabs)


UnKnoWn-Consortium:
Preview:

aldacron:
I wonder if we can right align the User ID, User created on, User class strings

UnKnoWn-Consortium:
@aldacron Done

Issue #18, opened 2017-05-13
Request: Implement a Torznab Compatible API


CodeMonkey626:
I would be great if nyaa.si could add a [Torznab](https://github.com/Sonarr/Sonarr/wiki/Implementing-a-Torznab-indexer) compatible API to query for torrents.

This would allow for direct integration with software such as [Sonarr](https://github.com/Sonarr/Sonarr) and [Radarr](https://github.com/Radarr/Radarr) as well ensuring applications don't need to resort to scraping to search for indexed torrents.

Background on Sonarr:

> Sonarr is a PVR for Usenet and BitTorrent users. It can monitor multiple RSS feeds for new episodes of your favorite shows and will grab, sort and rename them. It can also be configured to automatically upgrade the quality of files already downloaded when a better quality format becomes available.

Background on Torznab:

> The Newznab api specification is built around a simple rss feed with filtering and paging capabilities. It was designed for usenet primarily, but it can be easily adapted for torrents.

Torznab Documentation:
https://github.com/Sonarr/Sonarr/wiki/Implementing-a-Torznab-indexer

Taloth:
:) I was posting an issue as well: copy/paste and I'll close the dupe:

----

I'm one of the Sonarr developers and I've been keeping an eye on the fallout of nyaa closing down.
With it, a lot of our users have been asking for alternatives. With some working rss feeds, but lacking search capabilities.

For that I'd like to propose that nyaa.si supports a torznab api. In essence it's an enhanced rss feed with additional attributes and paging support as well as query parameters to specify search parameters.
The Torznab api specification is based on Newznab. The Newznab spec is used by +90% of all usenet indexers and has provided a standardized way for several applications to communicate with all those usenet indexers, saving the apps the trouble of implementing an api client for each site separately.
Torznab is intended to serve a similar purpose for torrent indexers.

With nyaa.si being open-source, I'm sure I could find some contributors willing to invest some time. But that depends on whether the nyaa devs want it.

Please let me know if you're interested.

PS: The most basic search capability is based on free text search. Which should be easy enough to implement for a contributor, but has certain performance implications. Eventually indexing by (for example anidb id) & absolute episode numbers could seriously improve the performance. So it's a matter of "how far do you want to go".

nyaadev:
I don't see a problem with adding this as far as it is possible.
Right now the rss search works exactly the same as the normal homepage search.
Adding the info we already have on the torrents to the rss is easy, but that's not a lot right now...

[...]

PR #108, opened 2017-05-17
Template refactoring + Torznab api proof of concept

Taloth:
Issue Ref https://github.com/nyaadevs/nyaa/issues/18

Showing how torznab would look. There are some side changes as well (yes, I know...).

Goal of the torznab api is to create a relatively consistent api surface for the various applications to utilize, such as Sonarr, SickRage/Gear, Medusa etc. Without each app having to implement an api for each site separately.

First, to get a bit more consistency in the template rendering. I wrapped the elastic search result with some objects that do the necessary conversions. It's not perfect, some things are still missing. But I'd appreciate if a python dev could look at that and hopefully turn it into something acceptable.
My main concern is that elastic misses a few bits of information, requiring the wrapper to query the db. But some stuff we can leave out, such as poster/uploader name. Saving a db query. Others, such as category name can be cached.
Either way, this cleans up most uses of 'use_elastic' out of the rss and torznab templates.

The commits are littered with TODO comments and I haven't run pep8 yet.
There are a couple of things todo with the various query params that torznab has. Primarily it not actually using the category atm.
Second is the 'quality filter', I was thinking about adding the query param 'tag' which could allow '!remake,trusted' for example. It's not a query param defined in the spec but it kinda makes sense and yet is generic enough.

But first I wanted you guys to have a chance to shoot at it, and hopefully inspire a python dev to take over the actual coding. (I'm not a python developer)

For those interested: On https://github.com/Taloth/nyaa/commit/342ab39b7fff01369b0ab5438d3861adb967cfee I've added a hacky but working docker-compose environment that spins up the stack in a set of docker containers (except for db initialization and elastic sync). It's not part of the PR.


CodeMonkey626:
Tagging @labrys who volunteered to review the PoC.

labrys:
@CodeMonkey626 What python versions are targeted and do you want style recommendations too such as PEP8 etc... also do you have a local style guide that you follow?

CodeMonkey626:
@labrys based on the [nyaa repo readme](https://github.com/nyaadevs/nyaa/blob/master/README.md), python 3.6.1, PEP8 and be sure to run lint.sh.

aldacron:
Since I'm responsible for the search part, I might have to take some time to review this to make sure nothing important breaks. But can you help us understand the value proposition of integrating this API? 

It might also help to separate out the refactoring from the addition of the API so we don't lump issues together.

CodeMonkey626:
@aldacron It would allow for direct and easy integration with multiple PVR tools like [Sonarr](https://github.com/Sonarr/Sonarr). The linked issue, #18, covers a lot of the value prop, but @Taloth may be able to expand additionally if needed.

nyaa/search.py
```diff
@@ -3,15 +3,99 @@
...
+    def __init__(self, elasticResult):
+        self.elastic = elasticResult
+
+    def __getattr__(self, name):
+        # Passthru for attributes that are the same
+        if name in self.knownAttributes:
+            return self.elastic[name]
+
+        # Safetynet: throw if an attribute is accessed which we don't implement    
+        raise Exception("Unknown attribute {}".format(name))
```
labrys: @Taloth Is it really necessary to only allow attributes in `knownAttributes`?

For example this method would be more flexible and more pythonic written as follows:

```python
def __getattr__(self, item):
    try:
        return self.elastic[item]
    except KeyError:
        # Raise an AttributeError if the key does not exist
        raise AttributeError(
            '{class_name!r} object has no attribute {attr!r}'.format(
                class_name=self.__class__.__name__,
                attr=item,
            )
        )
```
It raises a more appropriate exception (`AttributeError`) with a message that mimics the default python error message.  It allows for any attribute of `self.elastic` to be accessed directly as an attribute of `ElasticTorrentAccessor` thus allowing it to be flexible and extensible.

nyaa/search.py
```diff
@@ -3,15 +3,99 @@
...
+
+    @property
+    def created_time(self):
+        return datetime.strptime(self.elastic.created_time, '%Y-%m-%dT%H:%M:%S')
```
labrys: What's the purpose of all these `@property` attributes instead of just assigning them directly in the `__init__`?
```python
self.create_time = datetime.strptime(self.elastic.created_time, '%Y-%m-%dT%H:%M:%S')
```
They require evaluation on every call to the attribute, which only makes sense if `self.elastic` is expected to change between every call.  

If they are intended to be lazily evaluated, then the calculated value should be cached in a protected attribute.  An example for `created_time` follows:
```python
    @property
    def created_time(self):
        attribute = '_created_time'
        result = None
        try:
            result = getattr(self, attribute)
        except AttributeError:
            result = datetime.strptime(self.elastic.created_time, '%Y-%m-%dT%H:%M:%S')
            setattr(self, attribute, result)
        finally:
            return result
```

Taloth: For the simple properties I agree, for those doing a database looking it should use lazy loading unless they're guaranteed to be accessed. I'll look into that.

nyaa/search.py
Taloth: I wanted to cover the case when someone in the future adds something to the db, like created_time and adds them to the template, but forgets to handle it in the wrapper. created_time exists in both elastic and db, it's just a different type, so without the check it would fail silently and render an incompatible type in the ui.
But that hinges on someone else doing something wrong in the future. Your proposal is definitely lower maintenance.

Taloth:
@aldacron I didn't expect any of this to get merged in without parts of it being picked up by a dev for refactoring. That's why I rolled both into one PR.

The value proposition is: Quite a few apps already know how to talk to this api specification. The spec primary boils down to agreeing on a few standard query params and the format for the rss result so that any client can read it.
This is in contrast with every indexer/site having their own custom rss or api which would require the apps to implement the custom api for each site separately. And there are _considerably_ more sites than apps. :smile:

Right now nyaa is full-text search, which is fine and handled by the `q` query param. For tv series the spec normally uses `tvdbid`, `tvmazeid`, `season` and `ep`. which are simple indexer db lookups.
A couple of months ago I suggested adding `anidbid`/`aid` and `abs` to the list to do the same for the anime absolute format.
If this format can be agreed on, then any client that knows those query params could query nyaa (and similar sites) by increasingly efficient query params:
- q=Anime+Title+02  (full text)
- q=Anime+Title&abs=2  (db column & fulltext/series title)
- anidbid=1234&abs=2 (indexed db lookup)

If you have more questions about it, let me know. I didn't go into everything for the sake of brevity.

TL;DR: It's not intended to provide a full api for things like uploading, site control, etc. It's intended to standardize the rss and search api so any client can use any site.

nyaa/search.py
labrys: Well if you want a more defensive version, this would raise for undesired attributes:
```python
def __getattr__(self, item):
    if item not in supported_attributes:
        raise AttributeError(
            '{class_name!r} object does not support attribute {attr!r}'.format(
            class_name=self.__class__.__name__,
            attr=item,
        )
    else:
        try:
            return self.elastic[item]
        except KeyError:
            # Raise an AttributeError if the key does not exist
            raise AttributeError(
                '{class_name!r} object has no attribute {attr!r}'.format(
                    class_name=self.__class__.__name__,
                    attr=item,
                )
            )
```
But personally I'd prefer the following which raises a warning on unsupported attribute but allows execution to continue, or use logging instead and log each unsupported attribute access.
```python
import warnings

def __getattr__(self, item):
    try:
        result = self.elastic[item]
    except KeyError:
        # Raise an AttributeError if the key does not exist
        raise AttributeError(
            '{class_name!r} object has no attribute {attr!r}'.format(
                class_name=self.__class__.__name__,
                attr=item,
            )
        )
    else:
        if item not in self.known_attributes:
            warnings.warn('{class_name!r} object accessing unsupported attribute {attr!r}', AttributeError)
        return result
```

nyaa/search.py
```diff
@@ -3,23 +3,115 @@
...
 from nyaa import app, db
-from nyaa import models
+from nyaa import models, torrents
 
 import sqlalchemy_fulltext.modes as FullTextMode
 from sqlalchemy_fulltext import FullTextSearch
 from elasticsearch import Elasticsearch
 from elasticsearch_dsl import Search, Q
 
+class ElasticTorrentStatsAccessor:
```
labrys: What's the purpose behind having an `ElasticTorrentStatsAccessor` class, as its properties can be accessed from the `ElasticTorrentAccessor` directly (particularly with my recommended changes to `__getattr__`)

nyaa/routes.py
```diff
@@ -6,7 +6,8 @@
 from nyaa import torrents
 from nyaa import backend
 from nyaa import api_handler
-from nyaa.search import search_elastic, search_db
+from nyaa import api_torznab
+from nyaa.search import search
```
labrys: `search_db` is still used at [L311](https://github.com/Taloth/nyaa/blob/f569e8149db24118dda9a76e4c770ac093b01cf5/nyaa/routes.py#L311) but is no longer imported

nyaa/search.py
Taloth: Can it? it's a nested property: `torrent.stats.download_count`.
And if yes, wouldn't that be quite the hack? since getattr is already being used to forward property accesses to `torrent`.
Please note that the exposed property is `torrent.stats.download_count` while elastic has it in `elastic.download_count`.

nyaa/routes.py
Taloth: You're right. In my local copy I already eliminated some of that duplicate code, and uses search( exclusively.

Taloth:
@labrys @aldacron 
I've updated the code, included some of the python recommendations.

I opted to create a search_custom method that has no knowledge about rss or which view it's generating (UI vs business logic separation). Instead that behavior is defined by the other parameters. The search method is still there, but it's a wrapper, it for example forces the sorting and other parameters if it's for rss.
I've also removed a bit of duplicate code in routes, I tested what I could but I have only tested it on two torrents so not enough to truly test multiple pages.
Finally I added the notion of 'tags' to generalize the quality_filter. The search_custom method allows you to specify which tags to filter on. I did it primarily to keep stuff generic for torznab, but I'd imagine it could be useful in other cases.

I would like to recommend to include the category names and uploader names to elastic as well. That way the elastic search results can be used without additional db lookups.
The cost of the occasional 'reindex' in case the category name or uploader name changes is much lower than the additional db lookups that would be incurred every search.

I'd appreciate comments about the architectural changes and more importantly functionality.
Which steps need to be taken to get parts of this ready to be merged in? What decisions need to be made, and by who?
And, dare I ask: are there developers willing to pick it up and finalize it? (I'm just a Sonarr developer, with my own huge issue backlog, and not experienced with python at all)

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
-                'details', 'getnfo', 'get',
-                'cartadd', 'cartdel', 'comments', 'commentadd', 'user']:
+    elif mode in ['movie', 'music', 'book',
+                  'details', 'getnfo', 'get',
+                  'cartadd', 'cartdel', 'comments', 'commentadd', 'user']:
         return render_torznab_error(203)
 
-    return render_torznab_error(202)
+    else:
+        return render_torznab_error(202)
```
labrys: This many returns in a method is generally an antipattern.  It's better to store the result and return a single time at the end.

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
+    elif mode == 'register':
         return render_torznab_error(103)
 
     # modes specified in the newznab spec that are not supported.
-    if mode in ['movie', 'music', 'book',
-                'details', 'getnfo', 'get',
-                'cartadd', 'cartdel', 'comments', 'commentadd', 'user']:
+    elif mode in ['movie', 'music', 'book',
+                  'details', 'getnfo', 'get',
+                  'cartadd', 'cartdel', 'comments', 'commentadd', 'user']:
```
labrys: This list appears to be static.  If so then make it a module global constant to avoid recreating the list on every call to this function.

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
+
+def torznab_search(api_request):
+    page_size = TORZNAB_RESULTS_PER_PAGE
 
     # Default newznab query params, return empty result if unsupported to prevent clients from accessing the api as a traditional newznab indexer.
-    q = flask.request.args.get('q')
-    season = flask.request.args.get('season')
-    ep = flask.request.args.get('ep')
-    rid = flask.request.args.get('rid')
+    query_term = api_request.args.get('q')
```
labrys: The `query_` prefix is unnecessary here, and you don't use it consistently (tag, limit, etc are not prefixed below).

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
+    query_season = api_request.args.get('season')
+    query_episode = api_request.args.get('ep')
+    query_rageid = api_request.args.get('rid')
```
labrys: Since you are repeatedly getting items from `api_request.args` assign it to a variable a single time to avoid repeated look-ups:
```python
args = api_request.args
season = args.get('season')
episode = args.get('ep')
...
```

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
 
-    cat = flask.request.args.get('cat', default='100000')
+    if query_season is not None or query_episode is not None or query_rageid is not None:
```
labrys: The logic seems wrong here.  Do you really want it not have none of the following: season, episode, or rageid?

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
+    tag = api_request.args.get('tag', default='')
+    extended = api_request.args.get('extended', False, type=bool)
+    offset = api_request.args.get('offset', default=0, type=int)
+    limit = api_request.args.get('limit', default=page_size, type=int)
 
-    if season or ep or rid:
-        return render_torznab_error(201)
+    try:
+        limit = min(limit, page_size)
+        if limit == 0:
```
labrys: `if not limit`

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
+            raise EmptySearchError('Limit of 0 guaranteed guarantees empty result')
```
labrys: The error message doesn't make sense here.  Also generally raising an exception for a null set is undesirable.  Instead return an empty container (list, set, tuple, etc) since it evaluates falsey.  You can then just check `if not results`

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
+        if offset % limit:
+            app.logger.warn('offset {} isn\'t a multiple of the page size {}'.format(offset, limit))
```
labrys: Does it need to error if the offset isn't a multiple of the page size?  Couldn't you just start mid-page?  Maybe you mean to test that the offset is less than the page size?

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
+            'categories': categories,
+            'include_tags': include_tags,
+            'exclude_tags': exclude_tags,
+            'page': 1 + int(offset / limit),
+            'page_size': limit
+        }
 
-    query = search(**query_args)
+        query_results = search_custom(**query_args)
+    except EmptySearchError:
```
labrys: This `try: except` spans too many lines of code and is unnecessary.  If you just return when you realize the search is invalid, it greatly simplifies the code flow.

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
+def get_tag_filter(tag):
+    include_tags = []
+    exclude_tags = []
+    if tag:
+        for item in tag.split(','):
+            if item[0] == '!':
+                exclude_tags.append(item[1:].lower())
+            else:
+                include_tags.append(item.lower())
+    for incl in include_tags:
```
labrys: If you use sets instead of lists for 'include_tags' and 'exclude_tags'  this becomes a simple set operation instead of another iteration.
```python
if included.isdisjoint(excluded):
```

nyaa/routes.py
```diff
@@ -153,53 +154,15 @@ def home(rss):
...
-            pagination = Pagination(p=query_args['page'], per_page=per_page,
-                                    total=max_results, bs_version=3, page_parameter='p',
-                                    display_msg=SERACH_PAGINATE_DISPLAY_MSG)
-            return flask.render_template('home.html',
-                                         use_elastic=True,
-                                         pagination=pagination,
-                                         torrent_query=query_results,
-                                         search=query_args,
-                                         rss_filter=rss_query_string)
+    query_results = search(**query_args)
```
labrys: If you're going to just unpack the query args here, why not just use them here directly instead of building a dict for the args?
```python
results = search(
    term=term or '',
    user=user_id,
    sort=sort or 'id',
    ...
)
```

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
     return render_torznab_caps()
 
 
-def torznab_search():
-    page_size = app.config['RESULTS_PER_PAGE']
+def get_categories_filter(cat):
+    categories = []
+    if cat:
+        for item in cat.split(','):
+            # Parse 1_1 format for compatibility
```
labrys: Reduce the nesting here:
```python
items = cat.split(',') if cat else []
for item in items:
```

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
+        # Check if any of the specified categories 'stuck'.
+        if not categories:
+            raise EmptySearchError('Categories {} guaranteed to produce empty result'.format(cat))
+    return categories
+
+def get_tag_filter(tag):
+    include_tags = []
+    exclude_tags = []
+    if tag:
+        for item in tag.split(','):
```
labrys: Reduce the nesting here too, like above:
```python
items = tag.split(',') if tag else []
for item in items:
```

nyaa/search.py
labrys: `ElasticTorrentStatsAccessor` only exists to provide the `stats` property to `ElasticTorrentAccessor`.  The `stats` property is based off of `ElasticTorrentAccessor.elastic`.  With my change to `__getattr__` you could just do `ElasticTorrentAccessor.download_count` instead of `ElasticTorrentAccessor.stats.download_count`.

Now if you need to match an external interface dynamically, (not evident directly from the code with `ElasticTorrentAccessor` not being subclassed nor any comments to that effect), then that approach is fine, but should be well documented for its purpose.

nyaa/api_torznab.py
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
 
 
-def torznab_search():
-    page_size = app.config['RESULTS_PER_PAGE']
+def get_categories_filter(cat):
+    categories = []
+    if cat:
+        for item in cat.split(','):
+            # Parse 1_1 format for compatibility
+            cat_match = re.match(r'^(\d+)_(\d+)$', item)
```
labrys: You should compile this regex at the module level for a performance boost.
```python
# Just after module imports
CATEGORY_REGEX = re.compile(r'^(\d+)_(\d+)$')

# At actual usage point:
cat_match = CATEGORY_REGEX.match(item)

File: nyaa/api_torznab.py [hunk 19]
```diff
@@ -101,71 +112,113 @@ def torznab_caps():
...
+            if cat_match:
+                item = int(cat_match.group(1)) * 100 + int(cat_match.group(2))
+
+            num = int(item)
+            if num >= 100000:
+                categories.append(num - 100000)
+            elif num >= 1000:
+                categories.extend([x - 100000 for x in TORZNAB_CATEGORY_MAP.get(num, [])])
```
labrys: `categories` should be extended with a generator expression instead of a list comprehension to reduce the memory requirements of building an unnecessary list.  You can either do it directly in the extend as with your list comp, or create a genexpr variable and extend categories with it:
```python
categories.extend((x - 100000 for x in TORZNAB_CATEGORY_MAP.get(num, [])))
# OR
cat_map = (x - 100000 for x in TORZNAB_CATEGORY_MAP.get(num, []))
categories.extend(cat_map)
```

nyaa/search.py
```diff
@@ -55,38 +63,13 @@ def description(self):
...
-    @property
-    def sub_category(self):
-        # TODO: Same, hitting the db 
-        return models.SubCategory.by_category_ids(self.elastic.main_category_id, self.elastic.sub_category_id)
-
-    @property
-    def stats(self):
-        return ElasticTorrentStatsAccessor(self.elastic)
-
 class ElasticSearchResultAccessor:
```
labrys: It looks like you intend `ElasticSearchResultAccessor` to be a container of search results.  If so then you should extend an abstract base class.  For example:
```python
from collections import abc
class MyCustomSequence(abc.Sequence):
    def __init__(self, it):
        self.items = it
    
    def __len__(self):
        return len(self.items)
    
    def __getitem__(self, item):
        return self.items[item]
```

nyaa/search.py
```diff
@@ -55,38 +63,13 @@ def description(self):
...
 
     def __init__(self, elasticResult):
         self.elastic = elasticResult
+        self.total = self.elastic.hits.total
 
     @property
-    def total(self):
```
labrys: Extending an ABC, as mentioned above, would likely make this a good candidate for a `__len__` function.

nyaa/search.py
```diff
@@ -55,38 +63,13 @@ def description(self):
...
-        return self.elastic.hits.total
-   
-    @property
     def items(self):
```
labrys: This doesn't look like it has a purpose.

nyaa/api_torznab.py
Taloth: How will that work? Considering that an empty `cat` has special meaning. empty list vs list of unknown categories.
Which is why it's checked whether categories is empty inside the `if cat:`, coz if it's empty then the user did specify categories but they didn't translate to actual categories and an empty result set should be returned. If the user didn't specify a category, then it should search all categories.

nyaa/routes.py
Taloth: I didn't, that's existing code, and some query args are added/modified conditionally such as logged_in_user and admin.

nyaa/api_torznab.py
Taloth: The EmptySearchError is raised in various locations inside the search operation, including inside methods that are not part of the flow that returns the search result, and thus would not be able to do what you suggest.
Ironically I thought _this_ would be the more pythonic way of doing things, throwing the error and leave a higher up to decide whether to raise 404 or an empty result. If that's wrong, then fine, but please tell me how to do methods like `get_categories_filter`  and `get_tag_filter`.

nyaa/api_torznab.py
Taloth: Yes it throws when the offset isn't a multiple of the page size, that's what the warning literally says. Torznab allows you to set it mid-page, but it's rarely used.
Initially I had refactored search_elastic and search_db to use offset & limit instead of page & page_size. But that meant I couldn't use Paginator as was expected by parts of the UI code. And was getting increasingly complicated, I could've refactored that or have the legacy `search` create the paginator, but I felt that was more invasive and less likely to be accepted in the PR.
So instead I limited the torznab api to pages as 'acceptable' compromise.

nyaa/api_torznab.py
labrys: No you're right, this is ok, I actually meant to remove this comment when I saw the other sections of code you had raising `EmptySearchError`.

nyaa/api_torznab.py
Taloth: The original newznab spec didn't have 'supportedParams' in the caps, and assumed rid, season and episode query params were supported on the t=tvsearch call. If someone configures an old newznab client with the nyaa.si url you want to make sure it either returns an error or an empty list.
So yes, I don't want them coz nyaa.si doesn't support them, but for the above reason I also don't want to simply ignore them.

nyaa/api_torznab.py
Taloth: I think I did that coz pylint was complaining about some of those variable names.

nyaa/routes.py
labrys: For the conditional ones you evaluate them first then pass them.
```python
if flask.g.user and flask.g.user.is_admin:
    admin = True
else:
    admin = False
results = search(
    term=term or '',
    user=user_id,
    sort=sort or 'id',
    admin=
)
```
I realize that was existing code, but the comment still stands.  I would not recommend you change this though unless all such instances are changed.  Its better to be consistent.

nyaa/api_torznab.py
labrys: The code is functionally identical.  If cat is not falsey, then it gets split and assigned to items.  If it is falsey, then items becomes an empty list.  With an empty list `for item in items:` short circuits and skips the loop (same thing that `if cat:` accomplished).

nyaa/api_torznab.py
labrys: Probably for ep, since its short (i think pylint wants 3 character variables) but with episode it should be fine.

nyaa/api_torznab.py
labrys: Ok, that comment would probably be better served being located at the short-circuit logic, not the assignment.

nyaa/api_torznab.py
Taloth: I don't understand, yes it skips the loop, but doesn't skip the second if. The `if categories:` call inside the `if cat:` is functionally different if you omit the `if cat:`.
In the original code, if cat = '', then it will return []. If cat isn't empty, but results in an empty list (due to unknown category, for example), then it raises EmptySearchError.
In your suggestion, if cat = '', then it would raise EmptySearchError because that doesn't get short circuited.

Alternative: add as first (two) line(s) `if not cat: return []`, reduces the nesting but preserves the intended behavior, is that more acceptable?

labrys: Personally I was opting for treating no categories the same as no desired categories (see Principle of least surprise).

If I have list, and filter out any undesired items, that leaves me with a list containing only desired items.  If an empty list is considered acceptable, why shouldn't a list with all items removed (thus an empty list) also be acceptable?

Taloth: "Honey, can you go to the walmart pick something up from the fruits shelf, will ya?"
Guy comes home with some beer, chips and magazines... "Sorry honey, I swear, there was no fruits shelf so I assumed you didn't mind what I picked."

It _is_ the principle of least surprise, normally you would toss the specified categories in a db query, a simple many-to-many table that maps between category numbers and releases. Such a query would yield no results. And thus it shouldn't in this case either.

We're translating from external categories to internal ones, but from the user perspective the behavior should remain the same.

labrys: It's your implementation, so your choice, but this is essentially set arithmetic.  If you have a set with no items in it, that's the Null set.   You take a set with items, and remove all the items from it, what do you have left? Answer: The Null set.  For me that's the principle of least astonishment.

aldacron:
Shoot, this change seems really big and scary. @nyaadev thoughts?

xelra:
What's the status on this? Any progress?

There's probably a 100,000 people waiting for this. Maybe more. It would really be great if this could be merged.

aldacron:
well, it seems to my understanding that sonarr already supports nyaa.si. I don't think any of us wants to maintain this on our end -- can we shelve this until the codebase is more stable? we still have some features to add

Taloth:
TL;DR: There's serious technical debt in the db-business-ui layer separation that should be fixed, regardless of torznab.

@aldacron
The current codebase has logic that extends all the way from the db layer to the ui templates. where 'use_elastic' tags are used to apply different rendering (with the same result) depending on which database is used.
This is not a good design choice and makes adding features that much harder. ES was, quite literally, 'hacked in'.
Part of this PR contains suggestions on how to abstract that in the db layer, so that the business layer and ui layer has no awareness of what kind of db is underneath. This makes the UI templates smaller, easier to read and easier to maintain.
So it shouldn't be a question of 'we have features to add', but rather: a significant amount of technical debt was created on the search logic, and it should be fixed sooner rather than later.
I tacked this together having effectively no experience with python, so there are gonna be a lot of things not quite right yet, but by comparison it should be easy for you guys to pick that up and improve it.

Adding torznab should've been easy, couple of UI templates and a 200ish line api_torznab.py, but it requires a better search module.

I didn't ask for a code review, I didn't ask for it to be merged, I asked for someone to look at it functionally and architecturally. It's a proof of concept, showing the things that must be done to make this work.

I've mentioned earlier that the PR contains several TODOs that must be answered by you, and several design choices that must be decided upon. Yet the only response I've gotten was "Shoot, this change seems really big and scary." and some excellent python suggestions from labrys.
Quite frustrating considering I sunk several days labor into it.

So please, talk to me about this db problem. Read through the PR, look at the functional changes to search it tries to make and why, then let me know what you think about that.
I hope you recognize the value in keeping the db, business and UI layers separate whenever possible.

Ignore torznab for a sec. There are a few changes in the search layer inspired by torznab, such as the capability to search multiple categories, and filter by tags such as 'remake' arbitrarily instead of quality_filter, but those are details.
I much rather have you tackle the search layer separation and and leave torznab be for now.

If you can't spend an hour on a functional review, if you don't agree that the layer separation needs improving. Then by all means, close the PR, coz then we're not going to get anywhere.

aldacron:
thank you for your write-up and that makes a lot of sense. as I was the one who integrated ES, I definitely agree it was hacked in since we needed better search functionality _quickly_. 

thanks for the work done on this and I will take the time to hopefully dig deep into the topics and suggestions made here about the abstraction. however, that might be in a week or two as I'm preoccupied with other stuff.

we'll keep this PR open and hopefully use it as a reminder to tackle technical debt before it gets too deep. this project could've been planned much better, but we were also in a race against time so certain decisions are very sub-optimal.

once again, appreciate the work done here.

nyaa/api_torznab.py
pickfire: Why not `categaries.extend(x - 100000 for x in TORZNAB_CATEGORY_MAP.get(num, []))`?

xelra:
Has this api been implemented in another way now or why was it closed? Hundreds of thousands of people are eagerly awaiting this.

ReimuHakurei:
This PR was last touched a year ago, and the project codebase has changed substantially in that time. 

Nobody seems interested in completing this, and even if someone did, nobody is likely to want to maintain it either. 

Our current leech API is our RSS feed. This should be sufficient for use in all major torrent clients. 

Figuring none of our users have to my knowledge so much as mentioned this to us in many months, I don't think we have "hundreds of thousands" of users waiting on it.

ReimuHakurei:
Fair enough. 

I feel I should clarify that we aren't saying we won't ever implement this API but merely that we have other things that are more important right now, so things like this are comparatively low on our priority list. 

This particular PR had been abandoned for too long to be useful, and so was closed.

Taloth:
ReimuHakurei, did you read my and aldacron last reply? This PR was 90% refactoring of the db and business layers to make the whole stack easier to maintain, to deal with technical debt. The torznab api part is trivial and largely irrelevant.

Back then the PR was kept open so that aldacron could look at it and incorporate those changes.
That never happened, and the lack of separation between the db, business and ui layers still exist in the current codebase.

It's your right to close the PR for any reason, but please give me the courtesy by tell me tnx-but-no-tnx, instead of silently closing it.

As Sonarr dev, spending close to 40 hours on your project in a language I'm inexperienced with, this has been a rather disappointing collaboration.

Taloth:
I appreciate your follow up.

I indeed feel the 40h were wasted because nobody took the effort to review it, to say what they liked about the proposals and what they didn't. If that had happened, then those items could've been addressed individually as separate PRs.   
That's the frustrating part, because architectural improvements are meant to make implementing and maintaining features easier. And it should be everyone's concern to do so, not just aldacron.

For me this was an opportunity to make it easier for sites and apps to interface in a standardized manner by implementing torznab and was worth investing time in.

Anyway, thank you again for your reply.

Arylide:
It really just comes down to nobody wanting to work on this. I asked a few people, nobody wanted to do it. Maybe aldacron did originally, but that is seemingly a bygone thing now. I'm not really in the business of forcing people to work on stuff they don't want to. While it is true many things could be improved architecturally, the simple truth is the software "works" right now. And for a lot of people that is "good enough". Is that right? Is that good? I don't really know, but that is where we are on a lot of things.

Assuming the PR and branch stick around, the ideas and code aren't entirely wasted. Someone who is inclined to can come back and draw ideas from it. So I hope you don't feel that we don' appreciate your  your time. Many of the "core team" have a lot of other things going on in their lives and don't always want to work on this software, so progress is admittedly slow. That is just something we can't help as we're often in a "good enough" state for most people.

Issue #37, opened 2017-05-13
Change toggleDarkMode() so it loads a main-dark.css as well.


Ricardo1991:
Will allow to easily customize dark mode without changing the default bootstrap css files.

JSFernandes:
@Ricardo1991 what do you think of adding a `dark` class to either the `html` or `body` instead? This would prevent duplicate code for layout and etc between `main.css` and `main-dark.css`.

PR #111, opened 2017-05-18
Add site flavor and theme classes to body

Ricardo1991:
Not sure if the way I changed the theme class is the best way.

This will allow for easier manipulation of the stylesheet.

Related to #37


aldacron:
I feel like the <script> line has gotten waaay too long. We might want to factor it out into main.js or something along those lines.

PR #112, opened 2017-05-18
README: Fix typos

mc10:
Assuming the code doesn't have the same typos.


sharkykh:
Table prefiz one is actually outdated.
It's now called `SITE_FLAVOR` with the available options: `nyaa` or `sukebei`.

Don't know about the other one.

aldacron:
Yeah, we need to update our documentation. So much to do!

PR #115, opened 2017-05-18
Updated api/upload

kyamiko:
Updated api/upload to use code in backend.py

Cleaned up api_uploader.py

### DIFF
--- a/nyaa/backend.py
+++ b/nyaa/backend.py
@@ -26,7 +26,7 @@ def _replace_utf8_values(dict_or_list):
     return did_change
 
 
-def handle_torrent_upload(upload_form, uploading_user=None):
+def handle_torrent_upload(upload_form, uploading_user=None, fromAPI=False):
     torrent_data = upload_form.torrent_file.parsed_data
 
     # The torrent has been  validated and is safe to access with ['foo'] etc - all relevant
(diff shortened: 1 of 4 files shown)


nyaa/backend.py
```diff
@@ -26,7 +26,7 @@ def _replace_utf8_values(dict_or_list):
     return did_change
 
 
-def handle_torrent_upload(upload_form, uploading_user=None):
+def handle_torrent_upload(upload_form, uploading_user=None, fromAPI=False):
```
TheAMM: This does nothing?

nyaa/routes.py
```diff
@@ -696,8 +697,10 @@ def site_help():
 
 
 # #################################### API ROUTES ####################################
-# DISABLED FOR NOW
 @app.route('/api/upload', methods=['POST'])
 def api_upload():
-    api_response = api_handler.api_upload(flask.request)
+    is_valid_user, user, debug = api_handler.validate_user(flask.request)
+    if not is_valid_user:
+        return flask.make_response(flask.jsonify({"Failure": "Invalid username or password."}), 400)
```
TheAMM: I'd really prefer `failure` instead of `Failure` - this is an API, for computers. Lowercase keys!

utils/api_uploader.py
```diff
@@ -1,52 +1,115 @@
...
+    "display_name" : display_name,
+    "information"  : information,
+    "description"  : description,
+    "is_anonymous" : is_anonymous,
+    "is_hidden"    : is_hidden,
+    "is_remake"    : is_remake,
+    "is_complete"  : is_complete
+}
 
 files = {
```
TheAMM: Do we have to send everything as files...? There's no problem in sending variables and files at the same time in `requests.post(data=data)`.

Issue #131, opened 2017-05-18
Fix category name on ES search results


TheAMM:
Currently [the template is only using ids](https://github.com/nyaadevs/nyaa/blob/master/nyaa/templates/search_results.html#L55) for the category icon title, as seen here:
![nyaa_00 00 00-285x267-01](https://cloud.githubusercontent.com/assets/6820963/26227584/32d3bb3c-3c3c-11e7-9e23-ae25b541c476.png)

Pass a dict with the category names to the template or figure some other way to turn 1_2 into human readable name.

PR #135, opened 2017-05-19
Issue #131 category names don't appear

wfurine:
Issue #131 had the problem where elastic search caused category names not to appear.


nyaa/templates/search_results.html
```diff
@@ -52,7 +52,7 @@
 				{% set icon_dir = config.SITE_FLAVOR %}
 				<td style="padding:0 4px;">
 				{% if use_elastic %}
-				<a href="/?c={{ cat_id }}" title="{{ torrent.main_category_id }} - {{ torrent.sub_category_id }}">
+				<a href="/?c={{ cat_id }}" title="{{ torrent.main_category.name }} - {{ torrent.sub_category.name }}">
```
TheAMM: This doesn't actually work. The ES rows don't store the name, just the ID.
I know it's not easy to test the ES functionality as it requires a lot of setup, but if you can't test it (or aren't perfectly sure you know how it works), please leave it to those who have it set up.

PR #138, opened 2017-05-19
Add site flavor and darkmode to body class

Ricardo1991:
Retake on #127. It was kinda pointless to add a class saying the theme was light, so removed it.
Related to issue #37.

This will allow for easier manipulation of the stylesheet, as `<body>` will have "dark" as a class if dark mode is enabled.

addThemeClass() was added to main.js because the inline script can't touch `<body>` since it won't exist yet at that point.


nyaa/static/js/main.js
```diff
@@ -177,3 +177,9 @@ document.addEventListener("DOMContentLoaded", function() {
 // 		localStorage.setItem('theme', 'light');
 // 	}
 // }
+
+function addThemeClass(){
+	"dark" === localStorage.getItem("theme") && document.body.classList.add('dark')
+}
+
+window.onload=addThemeClass;
```
TheAMM: Don't replace `onload`, add a listener.

aldacron:
Can you resolve the conflict? Thanks, sorry

PR #146, opened 2017-05-20
Reporting functionality

nyaazi:
Custom report reason. Report actions: Close, Hide torrent, Delete torrent.

Might be a good idea to think of a way to notify uploaders of the taken action.


nyaazi:
Also, hope you won't mind the _fixed_ view.html. Tehe

TheAMM:
Screenshots would be nice, and I fear the "*fixed* view.html" will clash with #148.

nyaazi:
## The view page with report button, I such at design so someone proly should tweak this.
View page
## The report modal. Should be fine.
Report modal
## Reports page. 
Reports page

Also I "unfixed" the view.html so should be fine now.

nyaazi:
So I had to resolve conflicts once again(3rd ot 4th time) and I would like if we can go over the PR so you can express your concerns and suggestions.

PR #154, opened 2017-05-21
Expanded .gitignore

ipatrol:
Exactly what it says on the tin.


katnyaa:
Many of these entries are unrelated to the project (e.g. `.class`), and editor-specific ignores should go in individual developers' global gitignores (`git config --global core.excludesfile`).

PR #157, opened 2017-05-21
API Info v2

kyamiko:
Added API Info query.

As close to API Upload v2 behavior as possible.

### DIFF
--- a/migrations/versions/3001f79b7722_add_torrents.uploader_ip.py
+++ b/migrations/versions/3001f79b7722_add_torrents.uploader_ip.py
@@ -1,7 +1,7 @@
 """Add uploader_ip column to torrents table.
 
 Revision ID: 3001f79b7722
-Revises: 
+Revises:
 Create Date: 2017-05-21 18:01:35.472717
 
 """
--- a/migrations/versions/3001f79b7722_add_torrents.uploader_ip.py
+++ b/migrations/versions/3001f79b7722_add_torrents.uploader_ip.py
@@ -19,6 +19,7 @@ TABLE_PREFIXES = ('nyaa', 'sukebei')
 
 
 def upgrade():
+
     for prefix in TABLE_PREFIXES:
         op.add_column(prefix + '_torrents', sa.Column('uploader_ip', sa.Binary(), nullable=True))
         # ### end Alembic commands ###
--- a/nyaa/api_handler.py
+++ b/nyaa/api_handler.py
@@ -50,6 +50,22 @@ def api_require_user(f):
 
 # #################################### API ROUTES ####################################
 
+    if upload_request.method == 'POST' and form.validate():
+        torrent = backend.handle_torrent_upload(form, user, True)
+
+        return flask.make_response(flask.jsonify({'Success': int('{0}'.format(torrent.id))}), 200)
+    else:
+        return_error_messages = []
+        for error_name, error_messages in form.errors.items():
+            return_error_messages.extend(error_messages)
+
+        return flask.make_response(flask.jsonify({'Failure': return_error_messages}), 400)
+
+# ##################################### V2 BELOW #####################################
+
+
+# ###################################### UPLOAD ######################################
+
 # Map UploadForm fields to API keys
 UPLOAD_API_FORM_KEYMAP = {
     'torrent_file': 'torrent',
@@ -258,9 +274,11 @@ def ghetto_import():
 
     return 'success'
 
+
 # ####################################### INFO #######################################
 ID_PATTERN = '^[1-9][0-9]*$'
-INFO_HASH_PATTERN = '^[0-9a-fA-F]{40}$' # INFO_HASH as string
+INFO_HASH_PATTERN = '^[0-9a-fA-F]{40}$'  # INFO_HASH as string
+
 
 @api_blueprint.route('/v2/info/<torrent_id_or_hash>', methods=['GET'])
 @basic_auth_user
--- a/nyaa/api_handler.py
+++ b/nyaa/api_handler.py
@@ -284,20 +284,23 @@ INFO_HASH_PATTERN = '^[0-9a-fA-F]{40}$'  # INFO_HASH as string
 @basic_auth_user
 @api_require_user
 def v2_api_info(torrent_id_or_hash):
-    torrent = None
-    viewer = flask.g.user
+    torrent_id_or_hash = torrent_id_or_hash.lower().strip()
 
     matchID = re.match(ID_PATTERN, torrent_id_or_hash)
     matchHASH = re.match(INFO_HASH_PATTERN, torrent_id_or_hash)
 
+    torrent = None
+
     if matchID:
-        torrent = models.Torrent.by_id(torrent_id_or_hash)
+        torrent = models.Torrent.by_id(int(torrent_id_or_hash))
     elif matchHASH:
-        # Convert the string representation of a torrent hash into binary
+        # Convert the string representation of a torrent hash back into a binary representation
         a2b_hash = binascii.unhexlify(torrent_id_or_hash)
         torrent = models.Torrent.by_info_hash(a2b_hash)
     else:
         return flask.jsonify({'errors': ['Query was not a valid id or hash.']}), 400
+
+    viewer = flask.g.user
 
     if not torrent:
         return flask.jsonify({'errors': ['Query was not a valid id or hash.']}), 400
@@ -306,13 +309,13 @@ def v2_api_info(torrent_id_or_hash):
     if torrent.deleted and not (viewer and viewer.is_superadmin):
         return flask.jsonify({'errors': ['Query was not a valid id or hash.']}), 400
 
-    submitter = 'anonymous'
+    submitter = None
     if not torrent.anonymous and torrent.user:
         submitter = torrent.user.username
     if torrent.user and (viewer == torrent.user or viewer.is_moderator):
         submitter = torrent.user.username
 
-    files = ''
+    files = {}
     if torrent.filelist:
         files = json.loads(torrent.filelist.filelist_blob.decode('utf-8'))
 
@@ -322,10 +325,13 @@ def v2_api_info(torrent_id_or_hash):
         'url': flask.url_for('view_torrent', torrent_id=torrent.id, _external=True),
         'id': torrent.id,
         'name': torrent.display_name,
-        'hash': torrent.info_hash.hex(),
+        'hash_b32': torrent.info_hash_as_b32, #as used in magnet uri
+        'hash_hex': torrent.info_hash.hex(), #as shown in torrent client
         'magnet': torrent.magnet_uri,
         'main_category': torrent.main_category.name,
+        'main_category_id': torrent.main_category.id,
         'sub_category': torrent.sub_category.name,
+        'sub_category_id': torrent.sub_category.id,
         'information': torrent.information,
         'description': torrent.description,
         'stats': {'seeders': torrent.stats.seed_count, 'leechers': torrent.stats.leech_count, 'downloads': torrent.stats.download_count},
--- a/nyaa/api_handler.py
+++ b/nyaa/api_handler.py
@@ -280,7 +280,7 @@ ID_PATTERN = '^[1-9][0-9]*$'
 INFO_HASH_PATTERN = '^[0-9a-fA-F]{40}$'  # INFO_HASH as string
 
 
-@api_blueprint.route('/v2/info/<torrent_id_or_hash>', methods=['GET'])
+@api_blueprint.route('/info/<torrent_id_or_hash>', methods=['GET'])
 @basic_auth_user
 @api_require_user
 def v2_api_info(torrent_id_or_hash):
@@ -326,7 +326,7 @@ def v2_api_info(torrent_id_or_hash):
         'id': torrent.id,
         'name': torrent.display_name,
         'hash_b32': torrent.info_hash_as_b32, #as used in magnet uri
-        'hash_hex': torrent.info_hash.hex(), #as shown in torrent client
+        'hash_hex': torrent.info_hash_as_hex, #.hex(), #as shown in torrent client
         'magnet': torrent.magnet_uri,
         'main_category': torrent.main_category.name,
         'main_category_id': torrent.main_category.id,
@@ -335,7 +335,11 @@ def v2_api_info(torrent_id_or_hash):
         'information': torrent.information,
         'description': torrent.description,
         'stats': {'seeders': torrent.stats.seed_count, 'leechers': torrent.stats.leech_count, 'downloads': torrent.stats.download_count},
-        'filelist': files
+        'files': files,
+        # reduce torrent flags to 0/1
+        'is_trusted': 1 if torrent.trusted else 0,
+        'is_complete': 1 if torrent.complete else 0,
+        'is_remake': 1 if torrent.remake else 0
     }
 
     return flask.jsonify(torrent_metadata), 200
--- a/nyaa/api_handler.py
+++ b/nyaa/api_handler.py
@@ -325,8 +325,9 @@ def v2_api_info(torrent_id_or_hash):
         'url': flask.url_for('view_torrent', torrent_id=torrent.id, _external=True),
         'id': torrent.id,
         'name': torrent.display_name,
-        'hash_b32': torrent.info_hash_as_b32, #as used in magnet uri
-        'hash_hex': torrent.info_hash_as_hex, #.hex(), #as shown in torrent client
+        'creation_date': torrent.created_time.strftime('%Y-%m-%d %H:%M UTC'),
+        'hash_b32': torrent.info_hash_as_b32,  # as used in magnet uri
+        'hash_hex': torrent.info_hash_as_hex,  # .hex(), #as shown in torrent client
         'magnet': torrent.magnet_uri,
         'main_category': torrent.main_category.name,
         'main_category_id': torrent.main_category.id,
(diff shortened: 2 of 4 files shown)


nyaa/api_handler.py
```diff
@@ -186,3 +190,63 @@ def v2_api_upload():
...
+    if torrent.deleted and not (viewer and viewer.is_superadmin):
+        return flask.jsonify({'errors': ['Query was not a valid id or hash.']}), 400
+
+    submitter = 'anonymous'
+    if not torrent.anonymous and torrent.user:
+        submitter = torrent.user.username
+    if torrent.user and (viewer == torrent.user or viewer.is_moderator):
+        submitter = torrent.user.username
+
+    files = ''
```
TheAMM: Just default to a `{}` - mixing types is not nice for API users.

nyaa/api_handler.py
```diff
@@ -186,3 +190,63 @@ def v2_api_upload():
...
+@api_require_user
+def v2_api_info(torrent_id_or_hash):
+    torrent = None
+    viewer = flask.g.user
+
+    matchID = re.match(ID_PATTERN, torrent_id_or_hash)
+    matchHASH = re.match(INFO_HASH_PATTERN, torrent_id_or_hash)
+
+    if matchID:
+        torrent = models.Torrent.by_id(torrent_id_or_hash)
```
TheAMM: `torrent_id_or_hash` is a `str` here - `Torrent.id` is an `int`.

migrations/versions/3001f79b7722_add_torrents.uploader_ip.py
```diff
@@ -1,7 +1,7 @@
 """Add uploader_ip column to torrents table.
 
 Revision ID: 3001f79b7722
-Revises: 
+Revises:
```
TheAMM: Why is this touched?

nyaa/api_handler.py
```diff
@@ -186,3 +190,63 @@ def v2_api_upload():
...
+    if torrent.filelist:
+        files = json.loads(torrent.filelist.filelist_blob.decode('utf-8'))
+
+    # Create a response dict with relevant data
+    torrent_metadata = {
+        'submitter': submitter,
+        'url': flask.url_for('view_torrent', torrent_id=torrent.id, _external=True),
+        'id': torrent.id,
+        'name': torrent.display_name,
+        'hash': torrent.info_hash.hex(),
```
TheAMM: See https://github.com/nyaadevs/nyaa/blob/master/nyaa/models.py#L136

nyaa/api_handler.py
```diff
@@ -186,3 +190,63 @@ def v2_api_upload():
...
         mapped_errors = {UPLOAD_API_FORM_KEYMAP.get(k, k): v for k, v in upload_form.errors.items()}
         return flask.jsonify({'errors': mapped_errors}), 400
+
+
+# ####################################### INFO #######################################
+ID_PATTERN = '^[1-9][0-9]*$'
+INFO_HASH_PATTERN = '^[0-9a-fA-F]{40}$'  # INFO_HASH as string
+
+
+@api_blueprint.route('/v2/info/<torrent_id_or_hash>', methods=['GET'])
```
TheAMM: This doesn't have to be v2, does it? We never really had a proper v1 api set.

nyaa/api_handler.py
```diff
@@ -186,3 +190,63 @@ def v2_api_upload():
...
+        return flask.jsonify({'errors': ['Query was not a valid id or hash.']}), 400
+
+    if not torrent:
+        return flask.jsonify({'errors': ['Query was not a valid id or hash.']}), 400
+
+    # Only allow admins see deleted torrents
+    if torrent.deleted and not (viewer and viewer.is_superadmin):
+        return flask.jsonify({'errors': ['Query was not a valid id or hash.']}), 400
+
+    submitter = 'anonymous'
```
TheAMM: Since this is a name, I'd make it `Anonymous` - or possibly, since /user/Anonymous can exist, None

nyaa/api_handler.py
```diff
@@ -186,3 +190,63 @@ def v2_api_upload():
...
+    # Create a response dict with relevant data
+    torrent_metadata = {
+        'submitter': submitter,
+        'url': flask.url_for('view_torrent', torrent_id=torrent.id, _external=True),
+        'id': torrent.id,
+        'name': torrent.display_name,
+        'hash': torrent.info_hash.hex(),
+        'magnet': torrent.magnet_uri,
+        'main_category': torrent.main_category.name,
+        'sub_category': torrent.sub_category.name,
```
TheAMM: Could add the category IDs as well, separately or as N_N.

utils/api_info_v2.py
```diff
@@ -0,0 +1,85 @@
...
+conn_group.add_argument('-p', '--password', help='Password')
+conn_group.add_argument('--host', help='Select another api host (for debugging purposes)')
+
+resp_group = parser.add_argument_group('Response options')
+
+resp_group.add_argument('--raw', default=False, action='store_true', help='Print only raw response (JSON)')
+
+info_group = parser.add_argument_group('Info options')
+
+info_group.add_argument('-q', '--query', required=True, help='Torrent by id or hash Required.')
```
TheAMM: Use a positional for this, because there's nothing else you can input - the main point is the ID/hash, just like in upload you have to have a torrent.

utils/api_info_v2.py
```diff
@@ -0,0 +1,85 @@
...
+
+if __name__ == '__main__':
+    args = parser.parse_args()
+
+    # Use debug host from args or environment, if set
+    debug_host = args.host or os.getenv('NYAA_API_HOST')
+    api_host = (debug_host or (args.sukebei and SUKEBEI_HOST or NYAA_HOST)).rstrip('/')
+
+    api_query = args.query.lower().strip()
```
TheAMM: This is moot and if it has to be done, it should be done on the server.

utils/api_info_v2.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Verify query is either a valid id or valid hash
+    matchID = re.match(ID_PATTERN, api_query)
+    matchHASH = re.match(INFO_HASH_PATTERN, api_query)
+
+    if (not matchID) and (not matchHASH):
```
TheAMM: `if not (matchID or matchHASH):`

utils/api_info_v2.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Go!
+    r = requests.get(api_info_url, auth=auth)
+    response = None
+
+    if args.raw:
+        print(r.text)
+    else:
+        try:
+            response = r.json()
```
TheAMM: There's no point in parsing the JSON if we're not even prettyprinting it. Just print r.text.

nyaa/api_handler.py
kyamiko: No it doesn't. I tried to keep it as close to /v2/upload as possible.

utils/api_info_v2.py
```diff
@@ -28,13 +28,7 @@
...
 conn_group.add_argument('--host', help='Select another api host (for debugging purposes)')
 
-resp_group = parser.add_argument_group('Response options')
-
-resp_group.add_argument('--raw', default=False, action='store_true', help='Print only raw response (JSON)')
-
-info_group = parser.add_argument_group('Info options')
-
-info_group.add_argument('-q', '--query', required=True, help='Torrent by id or hash Required.')
+parser.add_argument('-q', '--query', required=True, help='Torrent by id or hash Required.')
```
TheAMM: This could still be just a `query` with a `metavar='ID_OR_HASH'` - it's the only required argument for the script, so don't make the user type a useless `-q`. It's clear enough what the argument is.

nyaa/api_handler.py
```diff
@@ -186,3 +190,69 @@ def v2_api_upload():
...
         mapped_errors = {UPLOAD_API_FORM_KEYMAP.get(k, k): v for k, v in upload_form.errors.items()}
         return flask.jsonify({'errors': mapped_errors}), 400
+
+
+# ####################################### INFO #######################################
+ID_PATTERN = '^[1-9][0-9]*$'
+INFO_HASH_PATTERN = '^[0-9a-fA-F]{40}$'  # INFO_HASH as string
+
+
+@api_blueprint.route('/v2/info/<torrent_id_or_hash>', methods=['GET'])
```
TheAMM: Make this just a `/info/`, no need for v2.

nyaa/api_handler.py
```diff
@@ -186,3 +190,69 @@ def v2_api_upload():
...
+        'hash_hex': torrent.info_hash.hex(), #as shown in torrent client
+        'magnet': torrent.magnet_uri,
+        'main_category': torrent.main_category.name,
+        'main_category_id': torrent.main_category.id,
+        'sub_category': torrent.sub_category.name,
+        'sub_category_id': torrent.sub_category.id,
+        'information': torrent.information,
+        'description': torrent.description,
+        'stats': {'seeders': torrent.stats.seed_count, 'leechers': torrent.stats.leech_count, 'downloads': torrent.stats.download_count},
+        'filelist': files
```
TheAMM: On another look, this could just be `files`, clean and simple - it's not really a list either.

nyaa/api_handler.py
```diff
@@ -186,3 +190,69 @@ def v2_api_upload():
...
+        files = json.loads(torrent.filelist.filelist_blob.decode('utf-8'))
+
+    # Create a response dict with relevant data
+    torrent_metadata = {
+        'submitter': submitter,
+        'url': flask.url_for('view_torrent', torrent_id=torrent.id, _external=True),
+        'id': torrent.id,
+        'name': torrent.display_name,
+        'hash_b32': torrent.info_hash_as_b32, #as used in magnet uri
+        'hash_hex': torrent.info_hash.hex(), #as shown in torrent client
```
TheAMM: We still have a property for this, but I suppose world doesn't end at that.

utils/api_info_v2.py
```diff
@@ -0,0 +1,63 @@
...
+import argparse
+import requests
+import json
+
+DEV_HOST = 'http://127.0.0.1:5500'
+NYAA_HOST = 'https://nyaa.si'
+SUKEBEI_HOST = 'https://sukebei.nyaa.si'
+
+API_BASE = '/api'
+API_INFO = API_BASE + '/v2/info'
```
TheAMM: Remember to change this as well if you change the route.

TheAMM:
Also, do rebase or otherwise resolve the merge conflict.

nyaa/api_handler.py
```diff
@@ -335,7 +335,11 @@ def v2_api_info(torrent_id_or_hash):
         'information': torrent.information,
         'description': torrent.description,
         'stats': {'seeders': torrent.stats.seed_count, 'leechers': torrent.stats.leech_count, 'downloads': torrent.stats.download_count},
-        'filelist': files
+        'files': files,
+        # reduce torrent flags to 0/1
```
TheAMM: Why? JSON can do boolean `true`/`false` just fine. And they *are* booleans.

utils/api_info_v2.py
```diff
@@ -60,4 +61,8 @@
     # Go!
     r = requests.get(api_info_url, auth=auth)
 
-    print(r.text)
+    #print(r.text)
+
+    # For printing unicode instead of unicode escape sequences
```
TheAMM: Is the output supposed to be JSON or human readable? `pprint` will be just rude to the user, since they can't use the raw JSON in another script; `pprint` does python-repr compatible output (generally).

It's better to implement `--raw` like the upload api to do `print(r.text)` and otherwise format the output into something like
```
#12345: 'Some Torrent Name' uploaded by 'Username' (123Mb)
  Anime / English-translated - (Complete, Trusted)
```
etc etc.
But, you know, on second thought maybe I'll do all that myself.

nyaa/api_handler.py
```diff
@@ -335,11 +335,12 @@ def v2_api_info(torrent_id_or_hash):
...
+        'filesize': torrent.filesize,
         'files': files,
-        # reduce torrent flags to 0/1
-        'is_trusted': 1 if torrent.trusted else 0,
-        'is_complete': 1 if torrent.complete else 0,
-        'is_remake': 1 if torrent.remake else 0
+        # reduce torrent flags to True/False
+        'is_trusted': True if torrent.trusted else False,
+        'is_complete': True if torrent.complete else False,
+        'is_remake': True if torrent.remake else False
```
shm0o: Just for the record, you could do a `bool(somevalue)` call for the same effect.

nyaa/api_handler.py
```diff
@@ -50,6 +50,22 @@ def decorator(*args, **kwargs):
...
+        torrent = backend.handle_torrent_upload(form, user, True)
+
+        return flask.make_response(flask.jsonify({'Success': int('{0}'.format(torrent.id))}), 200)
+    else:
+        return_error_messages = []
+        for error_name, error_messages in form.errors.items():
+            return_error_messages.extend(error_messages)
+
+        return flask.make_response(flask.jsonify({'Failure': return_error_messages}), 400)
+
```
shm0o: Is this intentional/correct? 'Cause the code is unreachable as is (see the `return` statement a few lines above for context).

nyaa/api_handler.py
```diff
@@ -335,11 +335,12 @@ def v2_api_info(torrent_id_or_hash):
         'information': torrent.information,
         'description': torrent.description,
         'stats': {'seeders': torrent.stats.seed_count, 'leechers': torrent.stats.leech_count, 'downloads': torrent.stats.download_count},
+        'filesize': torrent.filesize,
         'files': files,
-        # reduce torrent flags to 0/1
-        'is_trusted': 1 if torrent.trusted else 0,
-        'is_complete': 1 if torrent.complete else 0,
-        'is_remake': 1 if torrent.remake else 0
+        # reduce torrent flags to True/False
```
TheAMM: I cleaned up the properties and they will be booleans, so just do `torrent.trusted`.

utils/api_info_v2.py
```diff
@@ -61,8 +63,20 @@
...
+            print(r.text)
+            exit(1)
+
+        errors = rj.get('errors')
+
+        if errors:
+            print('Info request failed:',  errors)
+            exit(1)
+        else:
+            print("Torrent #{} '{}' uploaded by '{}' ({} bytes) {} / {}\n{}".format(rj['id'], rj['name'], rj['submitter'], rj['filesize'], rj['main_category'], rj['sub_category'], rj['magnet']))
```
TheAMM: Much better!

sharkykh: Well now, would you look at that... I didn't know that was possible.
You can refrence the object's elements from the format replacement fields instead of adding each element to the format function.
https://docs.python.org/2/library/string.html#format-string-syntax
```python
>>> t = {}
>>> t = {'one': 1, 'two': 2}
>>> '{0[one]}'.format(t)
'1'
>>>
```

migrations/versions/3001f79b7722_add_torrents.uploader_ip.py
sharkykh: @kyamiko  You should revert that change, IMO

nyaa/api_handler.py
sharkykh: The code is indented but doesn't seem to be in the correct function. I think this is a remanent of the V1 API route mistakenly re-added in a rebase/pull. But I might be wrong.

sharkykh:
Don't forget to check your code's lint. https://github.com/nyaadevs/nyaa#code-quality
 https://travis-ci.org/nyaadevs/nyaa/builds/237357111#L181

utils/api_info_v2.py
shm0o: Or even better, unpack the dictionary.
```python
>>> '{one}'.format(**{'one': 1, 'two': 2})
'1'
```

migrations/versions/3001f79b7722_add_torrents.uploader_ip.py
```diff
@@ -19,6 +19,7 @@
 
 
 def upgrade():
+
```
sharkykh: This too. There's just no reason for this file to be modified.

utils/api_info_v2.py
sharkykh: @shm0o 
I believe there's a `vformat` function (not sure about the name) so you can use it directly on the dictionary (no packing then unpacking)

TheAMM: On this PR, I'll take what I get.

nyaa/api_handler.py
shm0o: Yeah, that shouldn't be there in that case.

kyamiko:
The lint script only applies to files in ./nyaa, not ./utils. This should be fixed in a different commit.

utils/api_info_v2.py
```diff
@@ -0,0 +1,98 @@
...
+
+environment_epillog = '''You may also provide environment variables NYAA_API_HOST, NYAA_API_USERNAME and NYAA_API_PASSWORD for connection info.'''
+
+parser = argparse.ArgumentParser(
+    description='Query torrent info on Nyaa.si', epilog=environment_epillog)
+
+conn_group = parser.add_argument_group('Connection options')
+
+conn_group.add_argument('-s', '--sukebei', default=False,
+                        action='store_true', help='Upload to sukebei.nyaa.si')
```
TheAMM: `'Upload to sukebei.nyaa.si'` is wrong.

utils/api_info_v2.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
```
TheAMM: This is not a `v2` of the info api script, this is the first one. Plain `api_info.py` should be fine.

utils/api_info_v2.py
```diff
@@ -0,0 +1,98 @@
...
+conn_group = parser.add_argument_group('Connection options')
+
+conn_group.add_argument('-s', '--sukebei', default=False,
+                        action='store_true', help='Upload to sukebei.nyaa.si')
+
+conn_group.add_argument('-u', '--user', help='Username or email')
+conn_group.add_argument('-p', '--password', help='Password')
+conn_group.add_argument('--host', help='Select another api host (for debugging purposes)')
+
+parser.add_argument('-q', '--query', required=True, help='Torrent by id or hash Required.')
```
TheAMM: This should still be just a plain `parser.add_argument('hash_or_id', help='Torrent ID or hash (40-char hex string)')`

TheAMM:
Well, lessee if this squash will work out

PR #159, opened 2017-05-22
Test travis (ignore)

katnyaa:


sharkykh:
Error messages can also be formatted if needed, https://pycodestyle.readthedocs.io/en/latest/intro.html#example-usage-and-output

PR #162, opened 2017-05-22
Issue #128: Implemented static category link buttons

### LINKED ISSUE #128: Proposed Feature: Static Category Link Buttons (Remix)
There have been requests for the return of the static category link buttons in the old Nyaa (as in #106 and even in #92 ). I am proposing the following redesign that would be both modern and colorful :)

> 
> 
> 

Such design can also include a more complicated control flow. A small open button, which will reveal the subcategories upon clicks, can be added to the end each parent category link button. Though it may slightly duplicate the function provided by the search bar. 

P.S. The background colors are for reference only.

UnKnoWn-Consortium:
As suggested in #128 
fixes #128 

Scrollable when the screen width is insufficient

### DIFF
--- a/nyaa/forms.py
+++ b/nyaa/forms.py
@@ -1,3 +1,4 @@
+import flask
 from nyaa import db, app
 from nyaa.models import User
 from nyaa import bencode, utils, models
@@ -15,6 +16,7 @@ from wtforms.widgets import Select as SelectWidget
 from wtforms.widgets import html_params, HTMLString
 
 from flask_wtf.recaptcha import RecaptchaField
+from flask_wtf.recaptcha.validators import Recaptcha as RecaptchaValidator
 
 
 class Unique(object):
@@ -164,10 +166,6 @@ class EditForm(FlaskForm):
 
 
 class UploadForm(FlaskForm):
-
-    class Meta:
-        csrf = False
-
     torrent_file = FileField('Torrent file', [
         FileRequired()
     ])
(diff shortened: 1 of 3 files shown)


nyaa/templates/layout.html
```diff
@@ -44,8 +44,174 @@
 		<![endif]-->
 	</head>
 	<body>
+		{% set nyaa_cats = [
```
sharkykh: Just a comment...
I really think the categories should be loaded from the database when the server starts.
There is already a table for main_categories and sub_categories, they just need an extra column for `title`,
and a class that loads them then combines main & sub.
And last but not least, a function to load them lazily, as a list of dicts, like it's set up here.

sharkykh:
BTW, take a look at [Closing Issues via Pull Requests](https://github.com/blog/1506-closing-issues-via-pull-requests), it's pretty useful :)

nyaa/templates/layout.html
UnKnoWn-Consortium: I have the exact same feeling! It is pretty awkward to see these hardcoded into the templates. Moving it server-side can decouple the markup from the database. When there is a new category, we dun need to go through all the codes and apply the change manually. 

Even the `title` column is not needed as, currently, the title of the main categories are merely their name and those of the subcategories are a combination of the name of their respective parent category and their own name with "` - `" in between. It can be constructed on templates, unless this rule is just a coincident and may not be upheld in the future.

UnKnoWn-Consortium: How about this?

sharkykh: > Even the title column is not needed as, currently, the title of the main categories are merely their name and those of the subcategories are a combination of the name of their respective parent category and their own name with "`-`" in between. It can be constructed on templates, unless this rule is just a coincident and may not be upheld in the future.

Most, yes, but not all. `title` will still be needed.
For example, the title for `Anime / Anime Music Video` (1_1) is `Anime - AMV`.

> How about this?

I like the color, it matches the site's theme.
But maybe the other way around? So the blue color will mark the selected category, since it's bolder, it stands out more.

UnKnoWn-Consortium: Done~

shm0o:
Also squash all these merges commits.

PR #169, opened 2017-05-23
Add number of comments to torrent lists

sharkykh:
Partially related to #172.
Will only show when number of comments is above zero.

**Current preview:** // ([old](https://cloud.githubusercontent.com/assets/10238474/26455698/a986893a-4173-11e7-9868-4b84874c8fa3.png))
image

### DIFF
--- a/nyaa/search.py
+++ b/nyaa/search.py
@@ -25,6 +25,7 @@ def search_elastic(term='', user=None, sort='id', order='desc',
         'id': 'id',
         'size': 'filesize',
         # 'name': 'display_name',  # This is slow and buggy
+        'comments': 'comments_count',
         'seeders': 'seed_count',
         'leechers': 'leech_count',
         'downloads': 'download_count'
@@ -190,6 +191,7 @@ def search_db(term='', user=None, sort='id', order='desc', category='0_0',
         'size': models.Torrent.filesize,
         # Disable this because we disabled this in search_elastic, for the sake of consistency:
         # 'name': models.Torrent.display_name,
+        'comments': models.Torrent.comment_count,
         'seeders': models.Statistic.seed_count,
         'leechers': models.Statistic.leech_count,
         'downloads': models.Statistic.download_count
--- a/nyaa/static/css/main.css
+++ b/nyaa/static/css/main.css
@@ -62,6 +62,18 @@ table.torrent-list tbody tr td a:visited {
 	color: #1d4568;
 }
 
+table.torrent-list tbody .comments {
+	float: right;
+	border: 1px solid #d7d7d7;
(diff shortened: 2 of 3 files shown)


TheAMM:
I don't think this will work as-is, since the comment counts are not in the ES index (and therefore in the ES torrent instances). It'll take a bit of figuring to update the index cleanly and efficiently when the comment counts change.

sharkykh:
>I don't think this will work as-is, since the comment counts are not in the ES index (and therefore in the ES torrent instances). It'll take a bit of figuring to update the index cleanly and efficiently when the comment counts change.

@TheAMM 
Yep, that's what I was afraid of, as I can't test the elasticsearch locally right now. :disappointed:

aldacron:
Hmm, wonder how we can do this. Do we have a comment count somewhere? If we don't, this is going to get a little messy...

sharkykh:
@aldacron I have no clue when it comes to ES, but is it not possible to query the database for the amounts based on torrent IDs after search is completed?
Something like (not looking at the models, probably wrong)
```sql
SELECT COUNT * FROM `nyaa_comments` WHERE `torrent` IN (<comma-seperated-torrent-ids-from-search-results>)
```

TheAMM:
It is possible, but adds extra queries, especially since you'll have to `count(*)` all possible comments that are relationship'd to the torrent.
Spending the effort to sync up the counts to ES index would also give us sorting by comment counts.

TheAMM:
Heads up, you can use `torrent.comment_count` after #196 has been merged.

sharkykh:
@TheAMM 
So comments should get its own column instead of being inline like it is now?
If so, where do you think it fits best, after/before what column?

TheAMM:
Yeah, no own column, that right-aligned icon is very nice.

sharkykh:
@TheAMM 
I see you merged both of your PRs.

FYI, this PR should already be compatible with master and ready to be merged,
but I cannot verify ATM.
Of course, only **once all the torrents on the site have `comment_count` set**, like you said.

P.S. updated preview is available in the PR's description.

TheAMM:
Yeah, I know, I'm testing this atm

nyaa/search.py
```diff
@@ -25,6 +25,7 @@ def search_elastic(term='', user=None, sort='id', order='desc',
         'id': 'id',
         'size': 'filesize',
         # 'name': 'display_name',  # This is slow and buggy
+        'comments': 'comments_count',
```
TheAMM: `comment_count`, not `comments_count`!

sharkykh:
> Nevermind, impatient!

:laughing: :+1:

PR #178, opened 2017-05-24
Info hash search, hash on torrent page, template adjustments

### LINKED ISSUE #174: Allow search by infohash
Currently it is only possible to search by torrent name.

Just now I've found myself in a situation where I need to search by info hash (I have an old magnet link lingering in torrent client, that is definitely anime, but I have no idea what anime that is, since I've never found a peer via dht).

As a data point, you can search by infohash at acgnx.se.

TheAMM:
PR for image previews.

Fixes #174 

Torrent hash on torrent pages:
image

Searching for a hex representation of a hash as the sole input (eg. `5c39cf8ba80757a409d51bdacffab644f8a4960e`) will redirect to the torrent, if found:
image
(RSS or user page searches are not redirected)

Search field placeholder will show "Search for <user>'s torrents":
image

### DIFF
--- a/nyaa/models.py
+++ b/nyaa/models.py
@@ -205,6 +205,11 @@ class Torrent(db.Model):
     def by_info_hash(cls, info_hash):
         return cls.query.filter_by(info_hash=info_hash).first()
 
+    @classmethod
+    def by_info_hash_hex(cls, info_hash_hex):
+        info_hash_bytes = bytearray.fromhex(info_hash_hex)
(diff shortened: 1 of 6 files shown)


sharkykh:
It's a bit petty,
but I think you should drop the `s` when username ends with `s` as per English grammar.
For example:
`<prefix> HorribleSubs' torrents` and `<prefix> AnimeHunter's torrents`

PR #184, opened 2017-05-25
Darkened success and danger colors for the dark theme

thebombzen:
This should make the dark theme more usable


thebombzen:
Wow, okay, how did I miss that. Okay. Will do.

Ricardo1991:
Except you can't tell if the site is in darkmode without this #138

sharkykh:
@Ricardo1991 The `bootstrap-dark.min.css` file is loaded by [`setThemeDark()`](https://github.com/nyaadevs/nyaa/blob/cfa41bf3886395e157692e05ff6e226419bd0076/nyaa/static/js/main.js#L179), plus it also sets a localStorage item, `theme`, to `dark`/`light`.

Ricardo1991:
@sharkykh I know, but is altering bootstrap-dark.min.css the best option? Hence the css class so you can just put it on main.css

sharkykh:
@Ricardo1991 You're correct, I got confused. Obviously bootstrap shouldn't be altered.

ReimuHakurei:
What? In this case, altering Bootstrap is the proper thing to do, though obviously we need documentation of what exact changes were put into the Bootstrap customization tool to get this result so we can make further changes later.

thebombzen:
Success's background color was changed to #4c6442 (it was #dff0d8 previously). 
Danger's background color was changed to #988383 (it was #f2dede previously). 
A new <a> color for Success was added - #0d86ec, because the previous (default) <a> color on the new background was unreadable. 
A new <a> color for Danger was added - #105692, because the previous (default) <a> color on the new background was unreadable.

Tsunamical:
It'd be best if you added those to the commit message as well, that way we wouldn't need to dig back to the PR to find out *exactly* what was changed.

thebombzen:
I'll see if I can figure out how to work some git magic to make that happen

thebombzen:
Alright, I added it to the commit message and squashed everything together. It might be a bad idea to merge this commit technically because it seems that the proper way to do this would be to add a separate case to bootstrap's `dark.css` page which currently isn't there.

TheAMM:
#357

PR #66, opened 2017-05-15
Improve file list

sharkykh:
This is very, very simple, based on the table that currently exists.
Ideally, it should be done using `<li>` & `<ul>`, and allow for collapsing every directory tree (separately).
But I don't think I'm capable of doing that.

This configuration should be good in the meantime.

**Preview:**
image

PR #186, opened 2017-05-25
Basic collapsible file tree

sharkykh:
This was on my agenda since #66. @TheAMM 

**Notes:**
* I tried to keep jQuery/JS usage to the minimum.
* I also tried my best to make the [HTML source code look reasonable](https://cloud.githubusercontent.com/assets/10238474/26446389/772fc234-414c-11e7-95d0-96454b1310c8.png) with that Jinja2 loop...
* I don't feel like it's the best design, so suggestions are very welcome!
* Perhaps titles should be added to the anchors? "Collapse this folder", or something like that?

**Preview:** (3.3mb gif, sorry)
collapsiblefilelist

### DIFF
--- a/nyaa/static/css/main.css
+++ b/nyaa/static/css/main.css
@@ -293,15 +293,16 @@ a.text-purple:hover, a.text-purple:active, a.text-purple:focus { color: #a760e0;
 }
 
 .torrent-file-list ul {
-	padding: 5px 20px;
+	padding: 5px 20px 0px 20px;
 	list-style: none;
 }
 
-.torrent-file-list > ul:first-child {
+.torrent-file-list > ul {
 	padding: 0;
+	margin: 0;
 }
 
-.torrent-file-list li {
+.torrent-file-list li:not(:last-of-type) {
 	margin-bottom: 5px;
 }
 
--- a/nyaa/static/css/main.css
+++ b/nyaa/static/css/main.css
@@ -295,11 +295,18 @@ a.text-purple:hover, a.text-purple:active, a.text-purple:focus { color: #a760e0;
 .torrent-file-list ul {
 	padding: 5px 20px 0px 20px;
 	list-style: none;
+	display: none;
 }
 
 .torrent-file-list > ul {
+	display: block; /* First level always visible */
 	padding: 0;
 	margin: 0;
+}
+
+.torrent-file-list ul[data-show] {
+	/* Used to show first level's items based on amount */
+	display: block;
 }
 
 .torrent-file-list li:not(:last-of-type) {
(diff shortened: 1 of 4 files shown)


TheAMM:
Neat!

I'm a bit on the fence whether the directories should be pre-collapsed (as in all subitems as well), uncollapsed as is right now or collapsed when they contain more than N items (say, 10 or so), RFC, people?

The PR looks good but I'll take an hour or so to think about the collapsing a bit, while this is still contained in the PR context.

sharkykh:
@TheAMM 
I literally just had the same idea (to have all/some directories pre-collapsed, instead of the panel).
I'm exploring it right now. I'll post a preview image afterwards (without committing just yet).

Sn0wCrack:
I would say collapsing is a good idea for folders with larger contents is better than they're all precollapsed, some BD releases do only have like 4 or so entries. 

Perhaps up that limit to something like 13 however? Might make it easier for things like single cour batch releases as well.

ReimuHakurei:
For single cour batches, don't forget NCOPs and NCEDs. Maybe 20?

sharkykh:
@Sn0wCrack 
I was thinking in terms of level, depth and such.
So we'll always show the first level -
since every multi-file torrent contains a "base-folder" (see the preview gif)
And if there are less files than X (to be determined),
we show the folder's contents (no collapse) - but only one level down, meaning subfolders will still be collapsed.

If I'm making any sense, too hard for me to explain in words.

Sn0wCrack:
@sharkykh 
Limiting the first level may still need to happen, as some torrents may be batches for much longer shows, say 50+ long episode shows and such.

I do agree folders after the first level should start collapsed however, regardless of the length of their contents.

sharkykh:
I think this should be a no-brainer;
When you click to open the file list, you'll see this, regardless of the amount of files.
All the levels below are collapsed as well.

**Multi-file, with sub-folders:**
image

**Multi-file, no sub-folders:**
image

**Single file:**
image

sharkykh:
@ReimuHakurei 
I'm guessing these are like season/dvd packs of some sort? 20 sound good.
Latest commit implements that.

If we disable the collapsibility of the file list panel (which is ready to be committed), it really fits in.
Not too big, not too small. A little bit of scrolling didn't kill anybody yet.

If it annoys users, we can just add a small link somewhere under the torrent info: **Jump to comments**...

TheAMM:
Works for me

sharkykh:
@TheAMM Did you not want to disable the panel being collapsible?

TheAMM:
Oh, I missed that. Feel free to PR it again (or paste the changes).

PR #194, opened 2017-05-26
Url stuff

shm0o:
Per https://github.com/nyaadevs/nyaa/commit/1ab3d6e3eb1cc521e4f5c4ddd05a7c72f9d6bc5a#commitcomment-22282301.

### DIFF
--- a/nyaa/routes.py
+++ b/nyaa/routes.py
@@ -52,7 +52,7 @@ def redirect_url():
 
 
 @app.template_global()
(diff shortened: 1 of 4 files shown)


Mause:
Would it be worth using `functools.lru_cache` in `static_cachebuster`?

shm0o:
Doubt it, but if that were the case I think it should be done in another PR.

PR #201, opened 2017-05-27
Rewrite database models for declarative and update Alembic migrations

TheAMM:
Declarative base classes allows us to easily extend the bases into Nyaa and Sukebei models, with working FKs and relationships.
Both Nyaa and Sukebei tables will be created and defined regardless of set `SITE_FLAVOR` - this means we can use Alembic to generate and process database migrations pain-free.
This PR does not break the site: the `models.Torrent` etc will be chosen according to `SITE_FLAVOR`, being practically a perfect (but do see below) drop-in replacement for the old SQA models.

This rewrite does not change the actual schema; previously created databases should work fine - however, you may want to run `db_create.py` if your dev-database does not have the `sukebei_` tables.

*There is one possibly breaking change*: the [`uploader_flag_idx` is now prefixed](https://github.com/nyaadevs/nyaa/commit/a6826951faddd1e132c14c39895faa6684e4d4fc#diff-01da777df06353567b96d153bbf4a8e8R120) (because both flavors will always be created).
As far as I know this does not break the database nor SQA, but will possibly break migrations (as in downgrades or generating revisions from databases where `uploader_flag_idx` was not prefixed) - so rename your index or re-create the tables on your dev-database.


shm0o:
Nice. BTW, you could move the creation of the categories tables to the initial migration script you created here, so that there wouldn't be any need for `db_create.py` at all (through [alembic.operations.Operations.bulk_insert](http://alembic.zzzcomputing.com/en/latest/ops.html#alembic.operations.Operations.bulk_insert), for example).

sharkykh:
If categories are going to be worked upon, I suggest [actually loading the categories from the database on server init](https://github.com/nyaadevs/nyaa/pull/162#discussion_r117788959)

TheAMM:
@shm0o: `db.create_all()` is still cleaner than iterating through all database revisions. I had considered adding them in the migration, and that may happen later, but for now I'm content with `db_create.py`.

@sharkykh: Eventually!

shm0o:
Ah okay.

PR #204, opened 2017-05-28
Fix progressbar33 while fetching requirements

zaclimon:
It could also be possible that I'm the only one having this issue...

I'm using Debian Jessie for this.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ mysqlclient==1.3.10
 orderedset==2.0
 packaging==16.8
 passlib==1.7.1
-progressbar33==2.40.0
+progressbar33==2.4
 pycodestyle==2.3.1
 pycparser==2.17
 PyMySQL==0.7.11


shm0o:
Yeah, the version specifier is definitely mislabeled.

Though I'm wondering why was `progressbar2` removed in favor of that `progressbar33`. Giving a quick look at both projects and it seems that `progressbar33` development is quite dead at this point, and `progressbar2` is Python 2 and 3 compatible, so unless I'm missing something relevant there shouldn't be any real reason for the dependency change.

TheAMM:
The `progressbar2` PIP fetched for me didn't even install due to  what seemed like a bad `__init__` inside the setup. I did quick google early on, saw dozens of `progressbar`s and this one was suggested somewhere for compatibility for 3. I typo'd the version number, but otherwise the library works fine (and how much development do you need on a progress bar?)

Repeating record, but throw a(nother) PR with the change if you want to go back to `2`.

shm0o:
Hm that's weird then, guess I'd have to give it a quick check.

> (and how much development do you need on a progress bar?)

Well, yeah, not so much, I was just comparing the two of them.

shm0o:
Ok, so I just tried `progressbar2` and installation through pip seemed to work alright, and I also tried it from source and ran the unit tests and the examples and they worked all fine. I'm curious, what errors were you specifically getting?

Anyway, if `progressbar33` Just Works™ in this case then there's probably no need to go back to `progressbar2` for now.

TheAMM:
I'd tell you if I still had my terminals open. I tested `pip install` as well and it works now, so *shrug*.
Easier to stay put, though, like you said.

Issue #75, opened 2017-05-16
Adjust CSS for Sukebei to differentiate it from main site


TheAMM:
Just a simple recolor will do. Potentially add the `config['SITE_FLAVOR']` to an element high in the `layout.html` base and include some overrides for `body.sukebei <header>` etc.

decimoe:
Hot pink pls.

PR #206, opened 2017-05-28
75 Adjusts CSS for Sukebei

Rifu:
#75 
Modified layout.html to include the current site flavor so
it'll be easier to add flavor specific styles later.

I was unsure of how much we wanted Sukebei's styling to differ from the main site so I kept the changes to just the navbar. I was also unsure how we wanted to compress assets, so for the moment I just followed how main.css was inserted. I see we have flask_assets in the code base but it looks like we're not using it.

ss 2017-05-29 at 02 06 48
ss 2017-05-29 at 02 07 07

### DIFF
--- a/nyaa/static/css/sukebei.css
+++ b/nyaa/static/css/sukebei.css
@@ -1,0 +1,14 @@
+.sukebei .navbar{
+  background-color: #d81b60;
+  border-bottom: #a00037;
+}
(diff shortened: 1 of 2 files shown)


aldacron:
we'll be working on sukebei styling in the future, thanks for the suggestion though!

PR #211, opened 2017-05-29
Update the download buttons font size

nyaazi:
A somewhat hacky temporary fix untill a CSS master comes and makes something more permanent.

### DIFF
--- a/nyaa/templates/search_results.html
+++ b/nyaa/templates/search_results.html
@@ -42,10 +42,10 @@
 				{% endcall %}
 
 				{% if config.ENABLE_SHOW_STATS %}
-				{% call render_column_header("hdr-seeders", "width:50px;", center_text=True, sort_key="seeders", header_title="Seeders") %}
+				{% call render_column_header("hdr-seeders", "width:50px;", center_text=True, sort_key="seeders", header_title="Seeds") %}
 					<i class="fa fa-arrow-up" aria-hidden="true"></i>
 				{% endcall %}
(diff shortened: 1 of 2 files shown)


sharkykh:
Why not use a `btn-xs` class (smaller button) on the report button you added? Since it's the one that's too big?

Ricardo1991:
that actually seems the best

sharkykh:
@Ricardo1991 
It does, but I think it still makes the row's height a little bit bigger. Can you check with DevTools/FireBug?

nyaazi:
Because even before I thought the text is kinda small and the small report button looks silly to me,

Ricardo1991:
@sharkykh Yes, with the small button it is still 2px taller, from the button padding

ReimuHakurei:
Thoughts on this for the report button?

https://i.shoov.in/1496318594.png

sharkykh:
@ReimuHakurei I like that it's the same style as the download links.
As for the icon, http://fontawesome.io/icon/flag/ feels more fitting.

ReimuHakurei:
Ah, good point.

sharkykh:
The [(first) report button in the help template](https://github.com/nyaadevs/nyaa/blob/master/nyaa/templates/help.html#L40) should also be changed to match whatever change this PR will end up doing.

aldacron:
@ReimuHakurei I think fixing the report button should close this PR. The goal is to match the sizes, right?

ReimuHakurei:
Changed to btn-xs for now. Will consider better options in the future.

Issue #215, opened 2017-05-30
GFM Markdown Tables are not working in torrent descriptions


CounterPillow:
Github-flavoured Markdown tables are not being rendered in torrent descriptions.

Here's an example:

```
| No. | Title                             | Length |
| ---:|:--------------------------------- | ------:|
|  1  | Foo                               | 3:25   |  
|  2  | Bar                               | 3:35   |  
```

Should render as 

| No. | Title                             | Length |
| ---:|:--------------------------------- | ------:|
|  1  | Foo                               | 3:25   |  
|  2  | Bar                               | 3:35   |  

It does, however, result in plain-text output instead.

sharkykh:
The reason it doesn't work is because the site uses pure Markdown (CommonMark), and not GFM which is based on the CommonMark Markdown specification.

CounterPillow:
Yeah, but it would be nice if it did. I know this is non-trivial since it would require changing the markdown renderer.

sharkykh:
[Markdown-it](https://markdown-it.github.io/) supports it. But I don't know if the team wants to switch from pure CommonMark.

PR #218, opened 2017-05-30
Move to markdown-it

TheAMM:
As stated in #215, the Commonmark implementation we use is strictly commonmark and lacks (mostly) tables.
This PR replaces Commonmark with [markdown-it](https://github.com/markdown-it/markdown-it) and adds some CSS-fixes to the tables.
Elements tagged with `markdown-text` are parsed and rendered on `DOMContentLoaded` instead of separate `<script>` tags for every element (most notably with comments).

Preview:
image

Also escapes the description markdown in the editor just as is done on the torrent page (Cloudflare strips empty lines and in the editor, whitespaces is prepended to the lines)

### DIFF
--- a/nyaa/forms.py
+++ b/nyaa/forms.py
@@ -168,7 +168,7 @@ class EditForm(FlaskForm):
     information = StringField('Information', [
         Length(max=255, message='Information must be at most %(max)d characters long.')
     ])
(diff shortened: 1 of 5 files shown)


sharkykh:
Not sure, but isn't using `data-X="y"` attributes more valid, HTML-wise?

PR #230, opened 2017-06-01
Notifications

nyaazi:
Addition of notifications. 

Please review and give opinions. Especially on the table structure and the way mentions are handled.
Have to figure out a way to mark them as read in an efficiant way.

### DIFF
--- a/nyaa/models.py
+++ b/nyaa/models.py
@@ -475,9 +475,11 @@ class User(db.Model):
 
     nyaa_torrents = db.relationship('NyaaTorrent', back_populates='user', lazy='dynamic')
     nyaa_comments = db.relationship('NyaaComment', back_populates='user', lazy='dynamic')
+    nyaa_notifications = db.relationship('NyaaNotification', back_populates='user', lazy='dynamic')
 
     sukebei_torrents = db.relationship('SukebeiTorrent', back_populates='user', lazy='dynamic')
     sukebei_comments = db.relationship('SukebeiComment', back_populates='user', lazy='dynamic')
+    sukebei_notifications = db.relationship('SukebeiNotification', back_populates='user', lazy='dynamic')
 
     def __init__(self, username, email, password):
         self.username = username
@@ -558,6 +560,15 @@ class User(db.Model):
     @classmethod
     def by_username_or_email(cls, username_or_email):
         return cls.by_username(username_or_email) or cls.by_email(username_or_email)
+
+    @property
+    def unread_notifications_number(self):
+        flavor = app.config['SITE_FLAVOR']
+        if flavor == 'nyaa':
+            number = self.nyaa_notifications.count()
+        elif flavor == 'sukebei':
+            number = self.sukebei_notifications.count()
+        return number
 
     @property
     def is_moderator(self):
@@ -681,7 +692,8 @@ class NotificationBase(DeclarativeHelperBase):
 
     @declarative.declared_attr
     def user(cls):
-        return db.relationship('User', uselist=False, lazy="joined")
+        return db.relationship('User', uselist=False,
+                               back_populates=cls._table_prefix('notifications'), lazy="joined")
 
     def __init__(self, user_id, body):
         self.user_id = user_id
--- a/nyaa/models.py
+++ b/nyaa/models.py
@@ -724,7 +724,7 @@ class NotificationBase(DeclarativeHelperBase):
 
     @classmethod
     def get_notifications(cls, user_id, page):
-        notifications = cls.query.filter_by(read=False, user_id=user_id).paginate(page=page, per_page=20)
+        notifications = cls.query.filter_by(user_id=user_id).paginate(page=page, per_page=20)
         return notifications
 
     @classmethod
--- a/nyaa/models.py
+++ b/nyaa/models.py
@@ -570,9 +570,9 @@ class User(db.Model):
     def unread_notifications_number(self):
         flavor = app.config['SITE_FLAVOR']
         if flavor == 'nyaa':
-            number = self.nyaa_notifications.count()
+            number = self.nyaa_notifications.filter(Notification.read == False).count()
         elif flavor == 'sukebei':
-            number = self.sukebei_notifications.count()
+            number = self.sukebei_notifications.filter(Notification.read == False).count()
         return number
 
     @property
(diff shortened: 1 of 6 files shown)


nyaazi:
An image:

image

nyaazi:
Until someone can figure out a better way to mark as read, this will have to do.

aldacron:
What exactly is the point of notifications?

nyaazi:
@aldacron obviously to notify people. It's not very good when a torrent gets deleted/hidden and the uploader doesn't realize this and suddenly his torrent is just gone, right?
And it would be good if an user could mention the uploader or someone else to ask a question for example.

aldacron:
ah, I see, so it only applies to admin actions notifying users if their torrent gets moved/deleted or whatever? 

I'd think we need another set of eyes here like @nyaadev @TheAMM since they're better suited for DB changes

mreweilk:
This looks interesting but I don't think the reason should be included. Reason is whatever the reporter reports and might not actually be why it was deleted. We shouldn't mislead users.

nyaazi:
Yea, you have a point. Removed the reason.

sharkykh:
@nyaazi 
I just pushed a big change to master.
Are you okay with me rebasing your PR?

nyaazi:
No, go ahead.

nyaa/views/account.py
```diff
@@ -127,6 +127,27 @@ def profile():
...
 
+@bp.route('/notifications', endpoint='notifications', methods=['GET'])
+def view_notifications():
+    if not flask.g.user:
+        flask.abort(403)
+
+    page = flask.request.args.get('p', flask.request.args.get('offset', 1, int), int)
+    notifications = models.Notification.get_notifications(flask.g.user.id, page)
+    # Users unable to see which are read and unread (bolded)
+    # updates = models.Notification.mark_notifications_read(flask.g.user.id)
```
sharkykh: Have you tried rendering the template before calling `mark_notifications_read`?
I'm pretty sure that because you're altering the database, the notifications are marked as read before you actually render the template.
Try this:
```python
# [...]
notifications = models.Notification.get_notifications(flask.g.user.id, page)
template = flask.render_template('notifications.html', notifications=notifications)

updates = models.Notification.mark_notifications_read(flask.g.user.id)
db.session.commit()

return template
```

nyaa/views/admin.py
```diff
@@ -47,13 +47,23 @@ def view_reports():
                                  flask.url_for('torrents.view', torrent_id=torrent_id),
                                  report_user.username,
                                  flask.url_for('users.view_user', user_name=report_user.username))
+                notification = models.Notification(
+                    user_id=torrent.user.id,
+                    event='delete',
```
sharkykh: You can use `event=action,`

nyaa/views/admin.py
```diff
@@ -47,13 +47,23 @@ def view_reports():
...
             elif action == 'hide':
                 log = log.format(report_id, 'Hid', torrent_id,
                                  flask.url_for('torrents.view', torrent_id=torrent_id),
                                  report_user.username,
                                  flask.url_for('users.view_user', user_name=report_user.username))
                 torrent.hidden = True
                 report.status = 1
+                notification = models.Notification(
+                    user_id=torrent.user.id,
+                    event='hide',
```
sharkykh: You can use `event=action,` here as well

nyaa/views/torrents.py
```diff
@@ -40,7 +40,16 @@ def view_torrent(torrent_id):
 
         if comment_form.validate():
             comment_text = (comment_form.comment.data or '').strip()
-
+            mentioned = [s for s in comment_text.split() if '@' in s]
+            if len(mentioned) > 0:
```
sharkykh: 1. This won't work for users that have a space in their username (I think I saw at least some that does).
But that's not your issue. Your code works for all other cases (which is most of the users).

2. This is more Pythonic: `if mentioned:`
This means, if `mentioned` **isn't** Falsey (`False`, `None`, `empty string`, `empty list`, etc...)

nyaa/views/torrents.py
```diff
@@ -40,7 +40,16 @@ def view_torrent(torrent_id):
...
+                username = mentioned[0][1:]
```
sharkykh: It's worth mentioning is that you only use the first username in the comment.
I think you should use all of them.
Obviously we don't want any duplicates, so use something like:
```python
mentioned = [s[1:] for s in comment_text.split() if '@' in s]
# ...skipping...
for username in set(mentioned):  # sets only allow unique elements
# continue code...
```

nyaa/models.py
```diff
@@ -560,6 +569,15 @@ def by_username_or_email(cls, username_or_email):
         return cls.by_username(username_or_email) or cls.by_email(username_or_email)
 
     @property
+    def unread_notifications_number(self):
+        flavor = app.config['SITE_FLAVOR']
+        if flavor == 'nyaa':
+            number = self.nyaa_notifications.filter(Notification.read is False).count()
+        elif flavor == 'sukebei':
+            number = self.sukebei_notifications.filter(Notification.read is False).count()
+        return number
```
sharkykh: I can think of a more dynamic way for this:
```python
@property
def unread_notifications_number(self):
    notifications = getattr(self, '{}_notifications'.format(app.config['SITE_FLAVOR'].lower()))
    return notifications.filter(Notification.read is False).count()
```

nyaa/templates/layout.html
```diff
@@ -107,6 +107,15 @@
 						<li class="dropdown">
 							<a href="#" class="dropdown-toggle visible-lg visible-sm visible-xs" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
 								<i class="fa fa-user fa-fw"></i>
+								{% set notif_number = g.user.unread_notifications_number %}
+								{% if notif_number > 0 %}
```
sharkykh: The docs state that if the badge is not visible when empty, so there's no real need to check beforehand.
Docs: http://getbootstrap.com/components/#badges

(sorry, removed by mistake)

nyaa/templates/layout.html
```diff
@@ -134,6 +143,15 @@
 									</a>
 								</li>
 								<li>
+									<a href="{{ url_for('account.notifications') }}">
+										<i class="fa fa-envelope fa-fw"></i>
+										Notifications
+										{% if notif_number > 0 %}
```
sharkykh: Same as above (no need to check)

(sorry, removed by mistake)

mreweilk:
Any updates on this?

nyaazi:
I will submit a new PR in the next days.

PR #235, opened 2017-06-03
Document schema with diagrams

ipatrol:
Ok, I think I actually got the revision history clean this time.


.gitignore
```diff
@@ -1,10 +1,643 @@
+# If this isn't enough to cover all your frameworks, editors, and IDEs, then I have no idea what you are doing
```
shm0o: You already have been told in https://github.com/nyaadevs/nyaa/pull/154#issuecomment-303216953 that most of these don't apply to this project or are better suited for the developer's global configuration.

aldacron:
wait, what is the goal of this?

ipatrol:
Would you rather it be revised to include only software that is actually used by the project?

TheAMM:
@ipatrol The point is not to touch the `.git*` files at all.
Personally I don't think we need this either.

FichteFoll:
It's nice to have something to look at when you're not involved with the project but intend to, or are just a little interested, but it needs to be updated whenever the schema changes as well, as it would otherwise be misleading.

ipatrol:
Ok. Now only thing changed in the .gitignore is the addition of the virtualenv stuff, since the project instructions say to use it.

Arylide:
Seemingly abandoned PR, closing for now. If anyone wants to work on this please submit a new PR and reference this one.

PR #121, opened 2017-05-18
Issue #49: Add Torrent interface

nathancyam:
This adds a form wrapper when viewing one's own torrents. It identifies whether the user that you are viewing is you and then wraps the entire table in a form so you can change their properties.

I am still getting a grip on Python and Flask, so feedback would be very welcome.

PR #251, opened 2017-06-07
Test using Travis

sharkykh:
Based on @nathancyam's work from #121 

Had to drop Travis' container-based infrastructure and enable sudo, because some things are broken.
It was either a too old MySQL version to support `FULLTEXTSEARCH`, MariaDB not being able to install, or not being able to install `mysqlclient` module (Python) because it fails to build it...
It's not that big of a deal, it seems. Perhaps a little bit slower.

Also, a fix to the lint script's exit code (Travis builds always passing)

A new bash script with subcommands (`./dev.sh lint [-c]`,`./dev.sh test`) might be a good idea

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,29 @@ language: python
 
 python: "3.6"
 
-dist: xenial
-sudo: false
+dist: trusty
+sudo: required
 
 matrix:
   fast_finish: true
 
 cache: pip
 
-install: pip install --upgrade pycodestyle
+services:
+  mysql
 
-script: ./lint.sh --check
+before_install:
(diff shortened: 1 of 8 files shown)


aldacron:
I love you for this. will check it out in a bit.

sharkykh:
@aldacron :smile: 
I tried to start working on testing the API, but I'm still not very used to that.
I'm having a bit of a hard time, especially since I can't debug using PyCharm..
It's trial and error for me.

aldacron (commented):
fuck yeah testing

karolpodosek6-prog:
b6a818cdc6eabc75641905580a7a8683a7d6fe6b7d549ba885a43626f34515bf
Undress

PR #254, opened 2017-06-08
Load categories from DB

sharkykh:
Load the categories from the database for the search boxes in `layout.html`,
by adding `title` to the database, and using the [`get_category_id_map()`](https://github.com/sharkykh/nyaa/blob/50482340fb3696625053e7142b591aa61c0ba33d/nyaa/routes.py#L149-L168) cached function as a global template.
Note: DB changes in this PR.

* This **does not change** anything visibly in the front-end UI.
* **I advise to wait, and merge other `layout.html`-changing PRs before merging this.
_Especially #283 (merged) and #180_**

Additions:
- [x] `tabindent` template filter, that indents the code with tabs (only for `{{ code }}` statements)

**Update 2017-08-01:**
 * Rebased against master@[`87dd95f`](https://github.com/nyaadevs/nyaa/tree/87dd95f1e023416481a9c597b424830df6fe2614)
 * **Todo:** Update DB migration script to revise current head (ffd23e570f92)

### DIFF
--- a/db_create.py
+++ b/db_create.py
@@ -7,27 +7,87 @@ from nyaa.extensions import db
 app = create_app('config')
 
 NYAA_CATEGORIES = [
-    ('Anime', ['Anime Music Video', 'English-translated', 'Non-English-translated', 'Raw']),
-    ('Audio', ['Lossless', 'Lossy']),
-    ('Literature', ['English-translated', 'Non-English-translated', 'Raw']),
-    ('Live Action', ['English-translated', 'Idol/Promotional Video', 'Non-English-translated', 'Raw']),
-    ('Pictures', ['Graphics', 'Photos']),
-    ('Software', ['Applications', 'Games']),
+    {
+        'name': 'Anime',
+        'subcats': [
+            {'name': 'Anime Music Video', 'title': 'AMV'},
+            {'name': 'English-translated', 'title': 'English'},
+            {'name': 'Non-English-translated', 'title': 'Non-English'},
+            {'name': 'Raw'},
+        ]
+    },
+    {
+        'name': 'Audio',
+        'subcats': [
+            {'name': 'Lossless'},
+            {'name': 'Lossy'}
+        ]
+    },
+    {
+        'name': 'Literature',
+        'subcats': [
+            {'name': 'English-translated', 'title': 'English'},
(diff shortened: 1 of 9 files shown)


aldacron:
benefit of this change? also, the admin log PR has migrations as well. wonder if this will play well with that

sharkykh:
> benefit of this change?

Less crowded, a bit more readable `layout.html`, since
1. [I removed this part](https://github.com/nyaadevs/nyaa/blob/bd0f9edce3cfcc2606fbf530ddd66b00d32be0a6/nyaa/templates/layout.html#L163-L203).
2. Removed duplicate codes for the category `<option>`s generators by moving them into a macro (search `title="Category"` in the code linked above)

> also, the admin log PR has migrations as well. wonder if this will play well with that

I'm not taking any chances, :smile:
I'll just do another `./db_migrate.py migrate` and copy over my code.

aldacron:
k, maybe we can sit on this until the we resolve the previous PRs... I don't want to get too deep into dependency chasing :)

thanks though

sharkykh:
@aldacron 
Agreed, that's why I _also_ said we should wait :)

nathancyam:
Probably will need to consider how it would affect this: https://github.com/nyaadevs/nyaa/pull/121/files#diff-28a07da2ecde5c8715d3441469ddb927R143 once that gets pulled in. Still, I much prefer having the categories load rather than for us to hardcode some sort of notation.

sharkykh:
> Probably will need to consider how it would affect this: https://github.com/nyaadevs/nyaa/pull/121/files#diff-28a07da2ecde5c8715d3441469ddb927R143 once that gets pulled in. Still, I much prefer having the categories load rather than for us to hardcode some sort of notation.
> 
> Ahh, never mind. I can't read.

It's a good way of implementing it. Maybe I'll join the two methods, we'll see :)

***

Also, Oops!
I completely missed [`get_category_id_map()`](https://github.com/nyaadevs/nyaa/blob/bea63315cdf738fcdd3e644a74eb19d388594777/nyaa/routes.py#L149-L159), I'll take another look at it later.

Arylide:
Seemingly abandoned PR, closing for now. If anyone wants to work on this please submit a new PR and reference this one.

PR #257, opened 2017-06-09
Fix RFC822 filters + More tests

sharkykh:
# This PR is ready to be merged
Other than the changes to the rfc822 filters and the rename of the (unused) `flattenDict` function,
the changes are mostly contained in the `tests` folder.

***

* Make rfc822 filters compatible with Windows systems.
  > `.strftime()` is relative to the system it's being run on.
  > UNIX has `%s` for seconds since the EPOCH, Windows doesn't (raises `ValueError`).
  > Solution: use `.timestamp()` to achieve the same result on both/all platforms.
  > This also allows us to drop the float() around it, since it returns a float.
  [Docs for datetime.timestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp)
* I also took the opportunity to add tests for some of the filters in `routes.py`
* Changed `rfc822_es` filter's function name in order to be able to test it separately.
* Make 'tests' folder a Python package
  Now you can run tests with just `pytest tests`
* Update and organize `.gitignore` (deleted nothing, just added coverage files)
* Added tests for functions in the following files:
  - `routes.py`
  - `backend.py`
  - `bencode.py`
  - `utils.py`

### BTW, coverage! ~(updating as I go)~
The numbers differ a little in Travis because my database has torrents,
so when we test loading some pages like the home page and the RSS page,
it runs several functions from `torrents.py`
(The database on Travis doesn't have torrents or users just yet)

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,30 @@
-*.sql
-/.vscode
-/venv
-*.swp
+# Cache
 __pycache__
 /nyaa/static/.webassets-cache
+
+# Virtual Environments
+/venv
+
+# Coverage
+.coverage
+/htmlcov
+
+# Editos / IDEs
+/.vscode
+/.idea
+
+# Databases
+*.sql
 test.db
-install/*
+
+# Webserver
 uwsgi.sock
-/test_torrent_batch
+
+# Application
+install/*
 config.py
+/test_torrent_batch
 torrents
+
+# Other
+*.swp
--- a/.gitignore
+++ b/.gitignore
@@ -9,9 +9,8 @@ __pycache__
 .coverage
 /htmlcov
 
-# Editos / IDEs
+# Editors
 /.vscode
-/.idea
 
 # Databases
 *.sql
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -73,7 +73,15 @@ class TestFilters(NyaaTestCase):
             {'marco': 'polo'}, {},  # dictionaries
             None
         ]
-        self.assertEqual(filter_truthy(my_list), [item for item in my_list if item])
+        expected_result = [
+            True,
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -1,12 +1,30 @@
...
+/venv
+
+# Coverage
+.coverage
+/htmlcov
+
+# Editos / IDEs
+/.vscode
+/.idea
+
```
shm0o: IDEs' `gitignore`s are generally meant to (and preferably) be set on git's global configuration.

tests/test_filters.py
```diff
@@ -4,7 +4,7 @@
 from email.utils import formatdate
 
 from nyaa.routes import (_jinja2_filter_rfc822, _jinja2_filter_rfc822_es, get_utc_timestamp,
-                         get_display_time, timesince)
+                         get_display_time, timesince, filter_truthy, category_name)
```
shm0o: I know neither you nor this patch of code has anything to do with the `filter_truthy` code per se, but I thought I might bring this up now anyway since you're touching it. There's exactly zero need for `filter_truthy`, the Jinja2's `select` and `reject` filters without arguments serve literally the same purpose, a truth value test filtering on a sequence.

sharkykh: And `utils.flatten_dict()` isn't used since I changed the file list layout. :wink:
That's a good point, maybe in another PR though? I only tested the function. I do not want to edit templates and such in this PR.

.gitignore
sharkykh: Done. Also fixed that my annoying typo.

sharkykh:
I'm done for now, so this is ready to be merged whenever you'd like.

PR #266, opened 2017-06-17
Seeder/leecher styling

benwa:
Uses Bootstrap's contextual classes to color the seeder/leecher data instead of inline styles.

Before
screen shot 2017-06-17 at 09 52 30

After
screen shot 2017-06-17 at 09 56 18

### DIFF
--- a/nyaa/templates/view.html
+++ b/nyaa/templates/view.html
@@ -41,7 +41,7 @@
 			</div>
 
 			<div class="col-md-1">Seeders:</div>
-			<div class="col-md-5"><span style="color: green;">{% if config.ENABLE_SHOW_STATS %}{{ torrent.stats.seed_count }}{% else %}Coming soon{% endif %}</span></div>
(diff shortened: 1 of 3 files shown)


nyaa/templates/view.html
```diff
@@ -56,7 +56,7 @@ <h3 class="panel-title">
 			</div>
 
 			<div class="col-md-1">Leechers:</div>
-			<div class="col-md-5"><span style="color: red;">{% if config.ENABLE_SHOW_STATS %}{{ torrent.stats.leech_count }}{% else %}Coming soon{% endif %}</span></div>
+			<div class="col-md-5"><span class="text-success">{% if config.ENABLE_SHOW_STATS %}{{ torrent.stats.leech_count }}{% else %}Coming soon{% endif %}</span></div>
```
maddin77: Shouldn't this be `text-danger`?

benwa: Oops, you're right

sharkykh:
My opinion:
I don't like the colors, they're too dimmed.
And also, these don't work on the dark theme.

Arylide:
Same

PR #270, opened 2017-06-24
Update trackers.txt

pahakalle:
Added all the trackers zer0day.to says are stable.


ReimuHakurei:
This PR won't actually do anything, because Nyaa.si intentionally sets a cap on the maximum number of trackers that will be added to a torrent. 

At most, this would result in udp://tracker.leechers-paradise.org:6969/announce being added, because iirc the tracker number cap is currently set to 5.

Also, we don't have a lot of trackers for a reason; having a shitload of trackers causes torrent clients to take longer to do their initial announce, and DHT and Peer Exchange will do more than these added trackers would.

If you look back in the commit history, we did at one point have these trackers in trackers.txt, but intentionally removed them.

pahakalle:
Didn't even think about that.

PR #279, opened 2017-07-01
Create initial deploy via ansible

uMag:
Here is PR to remove all manual tasks to setup nyaa site, all that code is moved to ansible and ops-tools folde. And now you can setup nyaa via simple run of 
/ops-tools/orchestrate -i production -p nyaa.si  -r all "prepare,install,configure"
Python and working ssh to host is prerequsites


sharkykh:
The branch you worked on was outdated, you should rebase over the current master.

uMag:
@sharkykh Thanks, for the note, I've updated PR

mreweilk:
I'm not really sure this is something we're interested in.

uMag:
@mreweilk I think that will be useful for a new developers, or a way to spread nyaa deployments. 
Or i can create only deployment code repo.

ReimuHakurei:
Our goal is not to create a large number of Nyaa instances. Our goal is to create a torrent site. As such, that is outside the scope of this project. 

Either way, Nyaa is currently not hard to set up in any way, shape, or form. If someone couldn't figure out how to set it up, I'm not sure any code they would contribute would be up to our standards anyway.

uMag:
Ok, i'm understand feel free to reject or close PR

PR #284, opened 2017-07-05
Fix torrent deletion log

sharkykh:
* Fix [`url referenced before assignment.`](https://user-images.githubusercontent.com/10238474/27862507-2c946e02-618e-11e7-8e37-a26e9401201a.png)
* Fix every action logged as delete/undelete
* Use `editor` var (defined but not used) instead of `flask.g.user`

@mreweilk


mreweilk:
Doh! Thanks a bunch!

PR #292, opened 2017-07-14
[Restructure routes 2] Move account routes into blueprint

sharkykh:
**_Note:_ This PR is against the 'blueprints' branch.**
[2nd pull request out of 8 in total]

This PR moves the following routes into the `account` blueprint (`nyaa/views/account.py`):
* /login
* /logout
* /register
* /profile

`url_for()`-s in templates were updated,
and unused imports were removed from `routes.py`.

This also moves the following related functions/methods into the blueprint:
* `redirect_url()`
* `send_verification_email()`

Notes:
* `import config` was wrong. All usages of this import were changed to `app.config` (as it is throughout the application). It was only used by the email verification stuff.
* `@app.before_request` function uses the `logout()` route function, so the call was [changed to `views.account.logout()`](https://github.com/nyaadevs/nyaa/pull/292/files#diff-ac9cdeaf09e40bf3e4c4c70a2400e626R90)
* In the blueprint file, `get_activation_link()` is imported from routes.

***

You can also use git blame to see what code was actually changed by me (and what was plainly copied)
```console
git blame -C HEAD -- nyaa/views/account.py
```

### DIFF
--- a/nyaa/routes.py
+++ b/nyaa/routes.py
@@ -2,12 +2,8 @@ import json
 import math
 import os.path
 import re
-import smtplib
 from datetime import datetime, timedelta
-from email.mime.multipart import MIMEMultipart
-from email.mime.text import MIMEText
 from email.utils import formatdate
-from ipaddress import ip_address
 from urllib.parse import quote
 
 import flask
@@ -15,7 +11,6 @@ from flask_paginate import Pagination
 from werkzeug import url_encode
 from werkzeug.datastructures import CombinedMultiDict
 
-import config
 from itsdangerous import BadSignature, URLSafeSerializer
 from sqlalchemy.orm import joinedload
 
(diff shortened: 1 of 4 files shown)


sharkykh:
Team, is the `send_verification_email()` function used on the actual site?
This is probably the only thing I didn't test.

martstern:
@sharkykh It is not currently enabled, but it should work. If you want, you can easily test it with something like a gmail or other free email account if you plug in the credentials to use for it (the SMTP settings in config.py). Assuming it is still as originally written, with `USE_EMAIL_VERIFICATION` on, the newly created user will not be able to sign in until the verification link has been clicked.

sharkykh:
@martstern Thanks for the tip!
Seems to work. I mean, got the email, and the activation link worked.
I changed the user's status to INACTIVE through the database and logged out, and couldn't log back in until I visited the activation link again.

Worth mentioning that the current code treats inactive users like wrong username and/or password.
image
Might confuse users in the long run.

I'd say this is ready to be merged.

PR #300, opened 2017-07-20
[Restructure routes 4] Move user routes into blueprint

sharkykh:
**_Note:_ This PR is against the 'blueprints' branch.**
[4th pull request out of 8 in total]

This PR moves the following routes into the `users` blueprint (`nyaa/views/users.py`):

* /user/`<user_name>`
* /user/activate/`<payload>`

`url_for()`-s in templates and routes were updated.

This moves supporting functions and variables into other files:
* Into `nyaa.utils`:
  - `cached_function()`
  - `chain_get()`
* Into `nyaa.search`:
  - `DEFAULT_MAX_SEARCH_RESULT`
  - `DEFAULT_PER_PAGE`
  - `SERACH_PAGINATE_DISPLAY_MSG`
  - `_generate_query_string`
* Into the blueprint `nyaa.views.users`:
  - `get_serializer()`
  - `get_activation_link()`

**Notes:**
Some changes made to the master branch (https://github.com/nyaadevs/nyaa/commit/b51045503deab41e5db27caec725c9cd4b9c2f98) are not in this branch (to keep things simple).
When we open the PR from `blueprints` to `master`, it will cause conflicts, and I'll resolve them :+1: 

***

You can also use git blame to see what code was actually changed by me (and what was plainly copied)
```console
git blame -C HEAD -- nyaa/views/users.py
```

### Testing progress:
- [x] `get_serializer()`
- [x] `get_activation_link()`
- [x] GET `/user/<user_name>`
- [x] POST `/user/<user_name>` (change user level)
- [x] `/user/activate/<payload>`
- [x] `/adminlog`
- [x] GET `/reports`
- [x] POST `/reports` (perform action - close)
- [x] GET `/view/<torrent_id>/edit`
- [x] POST `/view/<torrent_id>/edit` (do edit)
- [x] GET `/view/<torrent_id>` (with comments)
- [x] POST `/view/<torrent_id>` (post comment)

PR #301, opened 2017-07-20
[Restructure routes 5] Move template filters and globals into blueprint

sharkykh:
# :white_check_mark: This PR should be squashed when merged.
**_Note:_ This PR is against the 'blueprints' branch.**
[5th pull request out of 8 in total]

This PR moves the following template filters and template globals
into the `template-utils` blueprint (`nyaa/template_utils.py`):

* Template filters:
  * `get_utc_timestamp` (named `utc_time`)
  * `get_display_time` (named `display_time`)
  * `_jinja2_filter_rfc822` (named `rfc822`)
  * `_jinja2_filter_rfc822_es` (named `rfc822_es`)
  * `timesince`
* Template globals:
  * `static_cachebuster` (including the `_static_cache` dict)
  * `modify_query`
  * `filter_truthy`

Notes:
* This PR is mainly a copy-paste of the code from `routes.py`. The only changes are:
  * Routed using `bp` instead of `app`
  * Changed a few docstrings to be encased in `"""` (triple double-quotes).
    PyCharm's inspection suggests using that instead of `'''` (triple single-quotes).
  * Fixed import line for [`url_encode` to import from `werkzeug.urls`](http://werkzeug.pocoo.org/docs/0.12/urls/#werkzeug.urls.url_encode).
* This blueprint injects everything in the app level, so there's no need to change any templates.
  So the fact that this file isn't in the `views` folder is intended.
* There is one template global left in `routes.py`: `category_name`.
  It will be moved in a later commit.
* This also updates the import line in the `tests/test_template_utils.py` file, so the tests could still pass.

***

You can also use git blame to see what code was actually changed by me (and what was plainly copied)
```console
git blame -C HEAD -- nyaa/template_utils.py
```

### Testing progress - render templates that use:
  - [x] `rfc822` filter (`rss.xml`)
  - [x] `timesince` filter (unused)
        -- it's now being done client-site with JS. The function is tested in `tests/test_template_utils.py`
            It should still work according to the code.
  - [x] `static_cachebuster` global (`layout.html`)
        -- I also add a test for the function in a later commit!
  - [x] `modify_query` global (`search_results.html`)
  - [x] `filter_truthy` global (`search_results.html`)
- **ElasticSearch only**: (I don't have ES, so can't verify, but should work according to the code)
  - [x] `utc_time` filter (`search_results.html`)
  - [x] `display_time` filter (`search_results.html`)
  - [x] `rfc822_es` filter (`rss.xml`)

### DIFF
--- a/nyaa/filters.py
+++ b/nyaa/filters.py
@@ -1,0 +1,107 @@
+import os.path
+from datetime import datetime
+from email.utils import formatdate
+
+import flask
+from werkzeug.urls import url_encode
+
+from nyaa import app
+
+bp = flask.Blueprint('filters', __name__)
+_static_cache = {}  # For static_cachebuster
+
+
+# ######################### TEMPLATE GLOBALS #########################
+
+@bp.app_template_global()
+def static_cachebuster(filename):
+    """ Adds a ?t=<mtime> cachebuster to the given path, if the file exists.
+        Results are cached in memory and persist until app restart! """
+    # Instead of timestamps, we could use commit hashes (we already load it in __init__)
+    # But that'd mean every static resource would get cache busted. This lets unchanged items
+    # stay in the cache.
+
+    if app.debug:
+        # Do not bust cache on debug (helps debugging)
+        return flask.url_for('static', filename=filename)
+
+    # Get file mtime if not already cached.
+    if filename not in _static_cache:
+        file_path = os.path.join(app.static_folder, filename)
+        file_mtime = None
+        if os.path.exists(file_path):
+            file_mtime = int(os.path.getmtime(file_path))
+
+        _static_cache[filename] = file_mtime
+
+    return flask.url_for('static', filename=filename, t=_static_cache[filename])
+
+
+@bp.app_template_global()
+def modify_query(**new_values):
+    args = flask.request.args.copy()
+
+    for key, value in new_values.items():
+        args[key] = value
+
+    return '{}?{}'.format(flask.request.path, url_encode(args))
+
+
+@bp.app_template_global()
+def filter_truthy(input_list):
+    """ Jinja2 can't into list comprehension so this is for
+        the search_results.html template """
+    return [item for item in input_list if item]
+
+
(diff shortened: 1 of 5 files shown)


sharkykh:
How do you feel about `filters.py` being outside the `nyaa/views/` folder?
My reasoning was because the defined filters are app-bound rather than actually blueprint bound, that would be more appropriate to put them in the app's root folder (`nyaa/`).

TheAMM:
I have no strong feelings either way.
The filters are relevant to the templates (so `nyaa/views` makes sense), *buut* they are not blueprints nor views.
If the file stays in `nyaa/`, I'd name it `[app/template]_filters.py` or something, although the simple `filters.py` meshes with `models`, `routes`, etc. *\*shrug[\*](https://mygi.ga/UmJ/adSfV.png)*

sharkykh:
@TheAMM I almost didn't catch the _shrug_ :stuck_out_tongue: 
I thank you for your input.
On a later commit, one towards making an app factory
```python
# (i.e.)
from nyaa import create_app
app = create_app('config')
```
I also move [this context processor](https://github.com/nyaadevs/nyaa/blob/fe4ac73ca294d6f4a49f01f4b9b7da7953fce939/nyaa/torrents.py#L90-L105) into this file (for app context reasons).

I could name this file `template_utils.py` or `templates.py`, or something like that...
It's actually better than `filters` because it contains more than just filters,
but still not the best, because future stuff might not be related to templates.

`filters.py` seemed close enough at the time of committing this PR.
before I knew about the context processor.

I know the naming is probably not _that_ important, but I'd like to get it right the first time.

TheAMM:
Ye, `template_utils.py`  in `nyaa/`, that's a good combination[.](https://mygi.ga/SQb/adRFV.jpg "I wanted to hide something else but didn't have anything relevant to my comments. Have some Python in anime instead. Maybe this dot is too small; a gift for the later curious generations.")

sharkykh:
Done testing (this PR), progress updated.
**But #300 needs to be merged first.**

PR #304, opened 2017-07-24
upload page: use alerts, add alert about rules

CounterPillow:
This changes the tracker announce URL message to be one of those bootstrap alert bubbles, and adds an additional warning which tells people to read the rules and not upload off-topic crap.

May not be to your liking since it takes up some space.
upload


Arylide:
Thanks for the PR, however this does indeed take up too much space, and is undissmissable, so it will be reverted at some point later. Also nobody reads the rules, even when shoved in their face, so it is probably just a moot addition to impede the uploader experience.

PR #308, opened 2017-07-25
bencode: fix nontermination on empty or malformed input

qqueue:
read(1) returns b'' on EOF, which wasn't checked for in a couple cases,
which could cause empty/truncated input to enter one of the `while True`
loops without hope of exit. programming is hard.

no tests, somebody else can do that.


qqueue:
>no tests, somebody else can do that.

fine, I've deigned to actually fix the tests.

PR #313, opened 2017-07-27
Update Gravatar

sharkykh:
* Update help page with Gravatar info.
* Use the rating parameter for Gravatar url generation.
  The default (when none is specified) is `G`.
   Limit images on Nyaa to `PG` and Sukebei to `X`
   See https://en.gravatar.com/site/implement/images/#rating


Arylide:
LGTM

PR #323, opened 2017-07-30
Update ES mapping with `asciifolding` filter

### LINKED ISSUE #321: Ignore characters like 'é'
On the old Nyaa.se if you searched for for example 'pokémon' you would get the same results as when you search for 'pokemon', could you implement that here too?

sharkykh:
Just from reading the [docs](https://www.elastic.co/guide/en/elasticsearch/guide/current/asciifolding-token-filter.html), unable to actually test... :<
(the line at EOF was automatically added by github)

Fixes #321

### DIFF
--- a/es_mapping.yml
+++ b/es_mapping.yml
@@ -12,6 +12,7 @@ settings:
         filter:
           - standard
           - lowercase
+          - asciifolding
       my_index_analyzer:
         type: custom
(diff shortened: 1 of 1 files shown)


sharkykh:
Thanks!
I'll close this for now, as this is not the correct solution.

PR #331, opened 2017-08-05
Expose soft delete to users and allow reuploading of deleted torrents. (fixes #39)

### LINKED ISSUE #39: Add a proper torrent deletion option
Currently deleted torrents are just hidden. 
I suggest a real delete and a infohash blacklist that will prevent said torrent from being uploaded again and a real delete that will allow upload again. I'm not sure why we would need to delete a torrent while still allowing it to be uploaded again but it should be an option.

If we are worried about rouge mods in the future perhaps true delete/blacklist should be reserved for admins.

Uploaders should also be able to delete their own torrents but perhaps it would be best to transparently set these as hidden in case an account has been hacked, though such behavior might lead users into thinking they will be able to upload the same torrent again which would not work.

nyaadev:
Add banning torrents for moderators which disallows reuploading.
New delete UI.
Fixes #39

### DIFF
--- a/nyaa/views/torrents.py
+++ b/nyaa/views/torrents.py
@@ -80,6 +80,7 @@ def view_torrent(torrent_id):
 def edit_torrent(torrent_id):
     torrent = models.Torrent.by_id(torrent_id)
     form = forms.EditForm(flask.request.form)
+    delete_form = forms.DeleteForm()
     form.category.choices = _create_upload_category_choices()
 
     editor = flask.g.user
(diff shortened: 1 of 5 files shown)


nyaa/views/torrents.py
```diff
@@ -147,9 +148,73 @@ def edit_torrent(torrent_id):
 
         return flask.render_template('edit.html',
                                      form=form,
+                                     delete_form=delete_form,
                                      torrent=torrent)
 
 
+@bp.route('/view/<int:torrent_id>/delete', endpoint='delete', methods=['POST'])
+def edit_torrent(torrent_id):
```
sharkykh: function name should be changed to something like `delete_torrent` or `soft_delete_torrent`

BotoX:
## Moderator views
### Not deleted torrent
screenshot_2017-08-05_20-26-12
### Deleted torrent
screenshot_2017-08-05_20-26-26
### Deleted and banned torrent
screenshot_2017-08-05_20-26-40

## User view
screenshot_2017-08-05_21-14-48

nyaadev:
## Moderator views
### Not deleted torrent
screenshot_2017-08-05_20-26-12
### Deleted torrent
screenshot_2017-08-05_20-26-26
### Deleted and banned torrent
screenshot_2017-08-05_20-26-40

## User view
screenshot_2017-08-05_21-14-48

PR #334, opened 2017-08-07
Separate main.css from light and dark and fix dark visited link colour

Golumpa:
Separated the main.css file to make it easier to make changes to either light or dark without it affecting both. Also changed the visited link colour in the dark main css to be a little bit lighter so it's easier to see with the dark background.

This is untested but I assume it will work.

### DIFF
--- a/nyaa/static/js/main.js
+++ b/nyaa/static/js/main.js
@@ -208,11 +208,13 @@ document.addEventListener("DOMContentLoaded", function() {
 
 // 	function setThemeDark() {
 // 		bsThemeLink.href = '/static/css/bootstrap-dark.min.css';
+// 		bsMainLink.href = '/static/css/main-dark.css';
 // 		localStorage.setItem('theme', 'dark');
 // 	}
 
 // 	function setThemeLight() {
 // 		bsThemeLink.href = '/static/css/bootstrap.min.css';
+// 		bsMainLink.href = '/static/css/main-light.css';
 // 		localStorage.setItem('theme', 'light');
 // 	}
 // }


sharkykh:
It seems like you made a copy of `main.css` and named it `main-dark.css`.

While this should work, I suggest something else.
For the sake of simplicity, and since the light theme is the default one, always import the light theme, and have the dark theme css override only certain styles, and have only stuff related to colors and icons.

This way, when changing the styling, you don't have to edit both files and make sure they have the same changes,
Instead, you change the big stuff in the default (light) css, and only change / add the needed bits on the dark css.

sharkykh:
Also, that small JavaScript in the layout file annoys me. Not that it's any of your doing, I just think there ought to be a better way to do this, one that doesn't involve minifed/uglified JS code within html.

benwa:
The JavaScript can be shrunken (but not removed entirely) using [alternative stylesheets](https://developer.mozilla.org/en-US/docs/Web/CSS/Alternative_style_sheets).
Roughly you'd have:

```html
<head>
<link rel="stylesheet" href="/static/css/bootstrap.min.css"> /* default bootstrap */
<link rel="alternate stylesheet" href="/static/css/dark.min.css" title="dark"> /* dark theme overrides */
<link rel="alternate stylesheet" href="/static/css/pink.min.css" title="pink"> /* pink theme overrides */

<script>
function switchTheme(theme) {
    $("link[rel='alternate stylesheet']").prop('disabled', true);
    if (theme)
        $("link[title='${theme}']").prop('disabled', false);
}
</script>
</head>
...
<footer>
<ul class="list-inline">
  <li><a href="switchTheme()">Default</a></li>
  <li><a href="switchTheme('dark')">Dark</a></li>
  <li><a href="switchTheme('pink')">Pink</a></li>
</ul>
</footer>
```

This can be optimized with more logic to prevent any flickering when changing themes and still needs to integrate the localStorage `theme` object to have it set on page load.

FichteFoll:
Quick question: Why aren't cookies considered for this instead of localStorage? At least for the initial rollout of the page.

I've seen the backrgound flicker a couple times when loading a new page while in Dark Mode, which can most likely be attributed to the stylesheets being swapped with JS while the page is loading. If the proper stylesheet (combination) was already in effect when the page is fetched from the server, this flickering would be gone.

Downside: Caching becomes almost impossible, although as far as I can see crsf tokens are already used anyway.

TheAMM:
@FichteFoll Caching, and because someone made it this way and nobody has touched the dark mode since.

Issue #109, opened 2017-05-17
[Search] Date filter


Kuraperunat:
Original nyaa had the ability to search given category torrents newer/older than n days and then sort the results.

Use case:
Searching for torrents posted past 1-30 days and sorting the results by download count was an useful way to browse popular stuff posted lately.

aldacron:
Probably need a better UI to surface this

nyaadev:
Needs HTML UI mock up.
Maybe an advanced search drop down like pantsu had it before?

UnKnoWn-Consortium:
This bootstrap favored date range picker can come in handy. 
[https://github.com/dangrossman/bootstrap-daterangepicker](url)
![2017-05-21_040653](https://cloud.githubusercontent.com/assets/9733856/26279068/9b4f9c8e-3ddc-11e7-887e-651000f08278.png)

For the design, I have conceived of two options. 

## **1. Primary Filter at the Search Bar**

It can be treated as a primary filter and added to the search bar, like this: 

![2017-05-21_041150](https://cloud.githubusercontent.com/assets/9733856/26279067/975b0802-3ddc-11e7-8802-c070d89df400.png)

**Advantages**

1. Easy to implement
2. Minimal change to the current layout

**Drawbacks**

1. Make the search bar more crowded
2. Need to open the menu in the mobile layout

## **2. Refinement filter above search results**

![2017-05-21_043947](https://cloud.githubusercontent.com/assets/9733856/26279189/90fba4dc-3ddf-11e7-827f-90edc12a1a5b.png)

**Advantages**

1. Search results can be refined directly
2. No need to overcrowd the search bar

**Drawbacks**

1. A larger change to the current layout

Kuraperunat:
This is still not implemented. There's a pull request by someone for this and I'm starving for this feature.

Anything going on?

Killualan:
+1

dreamboatcap:
This feature is requested a year ago. Any news?

wangsiw1:
3 months to 2 years. Really want this feature.

CounterPillow:
It's not in yet because we'd rather not have even more load on the database at the moment.

PR #338, opened 2017-08-16
Add date filter to search results

Azuny:
So I tried to implement date filter for search results like it was suggested in #109 . Basically just adding the calendar from https://github.com/dangrossman/bootstrap-daterangepicker, sending the date range as a request parameter and filtering directly in elastic, or in database model.

### DIFF
--- a/nyaa/templates/layout.html
+++ b/nyaa/templates/layout.html
@@ -48,6 +48,11 @@
 		<script src="{{ static_cachebuster('js/bootstrap-select.js') }}"></script>
 		<script src="{{ static_cachebuster('js/main.js') }}"></script>
 
+		<!-- Include Date Range Picker -->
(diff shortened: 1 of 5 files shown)


nyaa/templates/layout.html
```diff
@@ -48,6 +48,11 @@
 		<script src="{{ static_cachebuster('js/bootstrap-select.js') }}"></script>
 		<script src="{{ static_cachebuster('js/main.js') }}"></script>
 
+		<!-- Include Date Range Picker -->
+		<script type="text/javascript" src="//cdn.jsdelivr.net/momentjs/latest/moment.min.js"></script>
```
CounterPillow: please use cloudflare's CDNJS with an integrity check, no need to contact even more domains

ArjixWasTaken:
moment is too big of a dependency, you should look into dayjs

PR #340, opened 2017-08-19
help page: Add help for deletion

CounterPillow:
Some people may not realise that they can do this themselves, and maybe they'll consult the help page before abusing the report button.


Arylide:
>Expecting people to read the help page

PR #342, opened 2017-08-20
Minimum size for user-less anonymous torrents (+ post-WTForms validation)

TheAMM:
`MINIMUM_ANONYMOUS_TORRENT_SIZE` can be used to require a minimum total size of torrents uploaded by anonymous users (ie. without accounts) to cut down on spam.

As growing the original upload form validator too big is unreasonable, this PR also prepares for various "advanced" validations, such as blacklisting extensions per torrent category. In this post-WTForms context we have clean access to the database as well.
Torrents will be validated just before the final `session.commit()`, and any code calling `backend.handle_torrent_upload` should prepare for raised `TorrentExtraValidationException`s.

Any errors in the new validator will be added to the form, meshing nicely with the current error display:
image

### DIFF
--- a/nyaa/backend.py
+++ b/nyaa/backend.py
@@ -15,6 +15,11 @@ from nyaa.extensions import db
 app = flask.current_app
 
 
+class TorrentExtraValidationException(Exception):
+    def __init__(self, errors):
+        self.errors = errors
+
+
 @utils.cached_function
 def get_category_id_map():
     ''' Reads database for categories and turns them into a dict with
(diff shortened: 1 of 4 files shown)


Arylide:
¯\_(ツ)_/¯

nyaa/backend.py
```diff
@@ -44,7 +49,37 @@ def _replace_utf8_values(dict_or_list):
...
+    errors = {
+        'torrent_file': []
+    }
+
+    # Encorce minimum size for userless uploads
+    minimum_anonymous_torrent_size = app.config['MINIMUM_ANONYMOUS_TORRENT_SIZE']
+    if torrent.user is None and torrent.filesize < minimum_anonymous_torrent_size:
+        errors['torrent_file'].append('Torrent too small for an anonymous uploader')
+
+    print(errors)
```
sharkykh: @TheAMM missed one?

PR #354, opened 2017-09-01
Early filename validation

### LINKED ISSUE #330: Blacklist special characters from filenames
ie. Unicode overrides, which can be abused to spoof filenames in certain cases.
Potentially checking against other non-malicious but reserved characters such as `/` (and maybe even the ol' `?<>\:|"`)

TheAMM:
Currently just rejects torrents with RTL overrides in filenames.
Can be extended to reject uncommon (Windows-restricted) characters as well `?<>\:|"` and whatnot), or restricted (Windows) filenames (`NULL`, `COM`?)
image

Closes #330.

### DIFF
--- a/nyaa/backend.py
+++ b/nyaa/backend.py
@@ -49,6 +49,34 @@ def _replace_utf8_values(dict_or_list):
     return did_change
 
 
+def _recursive_dict_iterator(source):
+    ''' Iterates over a given dict, yielding (key, value) pairs,
+        recursing inside any dicts. '''
+    # TODO Make a proper dict-filetree walker
+    for key, value in source.items():
+        yield (key, value)
+
+        if isinstance(value, dict):
+            for kv in _recursive_dict_iterator(value):
+                yield kv
+
+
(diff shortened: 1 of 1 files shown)


ReimuHakurei:
I think filtering before displaying is a better idea than blocking uploads
altogether.

nyaa/backend.py
```diff
@@ -49,6 +49,34 @@ def _replace_utf8_values(dict_or_list):
...
 
+def _recursive_dict_iterator(source):
+    ''' Iterates over a given dict, yielding (key, value) pairs,
+        recursing inside any dicts. '''
+    # TODO Make a proper dict-filetree walker
+    for key, value in source.items():
+        yield (key, value)
+
+        if isinstance(value, dict):
+            for kv in _recursive_dict_iterator(value):
```
FichteFoll: `yield from` does this nicely, too.

TheAMM: Good point, I haven't really looked at all the three-isms.

PR #356, opened 2017-09-02
Read-only maintenance mode

TheAMM:
```
Disables all POSTs, optionally allowing users to log in (without updating last login date)
Blocked POSTs will redirect to the GET endpoint if possible, otherwise to referrer or in last case, home page.
API requests will get a plaintext message with 405 status code.
```
Every page will display a warning about the maintenance mode.
image

POSTs will get a redirect and `flask.flash()`
image

### DIFF
--- a/nyaa/views/account.py
+++ b/nyaa/views/account.py
@@ -21,6 +21,10 @@ def login():
 
     form = forms.LoginForm(flask.request.form)
     if flask.request.method == 'POST' and form.validate():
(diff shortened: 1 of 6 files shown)


nyaa/views/account.py
```diff
@@ -21,6 +21,10 @@ def login():
 
     form = forms.LoginForm(flask.request.form)
     if flask.request.method == 'POST' and form.validate():
+        if app.config['MAINTENANCE_MODE'] and not app.config['MAINTENANCE_MODE_LOGINS']:
```
TheAMM: I realized this is never reached. Oh well!

PR #360, opened 2017-09-06
Re-add uw0.xyz tracker (aka doko.moe)

Miiiiaaooooo:


Arylide:
The file permits a maximum of four trackers. Also it has been decided to stick with established and stable trackers.

Miiiiaaooooo:
I understand the fact that it only allows 4 trackers, sorry that i didn't know that before, however doko.moe tracker(now uw0.xyz) has been online for over 2 years and had quite a good uptime of around 99.5% during the last 2 years. I believe that is stable.

Arylide:
The uptime is currently reported as barely 85% percent, and I've observed multiple instances of downtime in the last few months. The domain has also been changed for seemingly no perceivable reason, the entire point of an open tracker is to be continually available, changing the domain name is the same as indefinite "downtime" when they eventually discontinue the redirection.

Miiiiaaooooo:
Since i changed domain i asked newtrackon and https://uptime.1mb.info/ to change them to the new one and thus they deleted the old tracker and added the new one, since we had a few hours downtime not long ago due to some problems with the hosting provider which cause the uptime to go down a lot since it technically was considered a new tracker, the connection wouldn't eventually be discontinued, i just changed domain for the sake of keeping doko.moe separate from the tracker, like i mentioned on your irc channel both links will work, but from now on the official link is tracker.uw0.xyz

Miiiiaaooooo:
I don't mind not having the tracker included, this is just related to the fact that you mentioned the tracker was not stable and it definitely is.

Arylide:
Noted.

Miiiiaaooooo:
The first one already explained, the second one im going to suppose its bugged since it shows its never been online and you can check its online right now.

Arylide:
Your decision to change the url is still strange and disturbing to me. I will stick with more established and reliable options, thank you for your contribution.

PR #377, opened 2017-10-06
Implement torrent nuking ability for mods

CounterPillow:
This deletes and bans all torrents of a specific user.

A current caveat is that it will delete both sukebei and nyaa torrents, but will only leave a log entry in the current flavour's log.

Also did some bootstrap untangling on the user view page.

Please review these changes carefully, it's my first time messing with the DB, and I did not check whether it sends the ban command to the tracker properly.

Also, we may want to restrict nuking to only supermods.

### DIFF
--- a/nyaa/views/users.py
+++ b/nyaa/views/users.py
@@ -6,11 +6,12 @@ from flask_paginate import Pagination
 
 from itsdangerous import BadSignature, URLSafeSerializer
 
-from nyaa import forms, models
+from nyaa import backend, forms, models
 from nyaa.extensions import db
 from nyaa.search import (DEFAULT_MAX_SEARCH_RESULT, DEFAULT_PER_PAGE, SERACH_PAGINATE_DISPLAY_MSG,
                          _generate_query_string, search_db, search_elastic)
 from nyaa.utils import chain_get
+from itertools import chain
 
 app = flask.current_app
 bp = flask.Blueprint('users', __name__)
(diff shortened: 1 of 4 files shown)


nyaa/views/users.py
```diff
@@ -100,6 +101,23 @@ def view_user(user_name):
...
+        num_banned = 0
+        for t in chain(user.nyaa_torrents, user.sukebei_torrents):
+            t.deleted = True
+            t.banned = True
+            backend.tracker_api([t.info_hash], 'ban')
+            db.session.add(t)
+            num_banned += 1
+
+        log = "Nuked {0} torrents of {1}".format(num_banned, user.username)
+        adminlog = models.AdminLog(log=log, admin_id=flask.g.user.id)
```
TheAMM:
```py3
for log_class in (models.NyaaAdminLog, models.SukebeiAdminLog): # Check these because I'm writing this from memory
    adminlog = ...
    db.session.add(adminlog)
db.session.commit()
```
to add logs to both flavors. You could also count them separately (`Nuked x nyaa, y suke torrents`)

nyaa/views/users.py
```diff
@@ -100,6 +101,23 @@ def view_user(user_name):
...
         return flask.redirect(url)
 
+    if flask.request.method == 'POST' and ban_form and ban_form.nuke.data:
+        num_banned = 0
+        for t in chain(user.nyaa_torrents, user.sukebei_torrents):
+            t.deleted = True
+            t.banned = True
+            backend.tracker_api([t.info_hash], 'ban')
+            db.session.add(t)
+            num_banned += 1
```
TheAMM:
```py3
nyaa_nuked = 0
suke_nuked = 0
# ...
for t in ...:
    if isinstance(t, models.NyaaTorrent):
        nyaa_banned += 1
    else:
        suke_banned += 1
```
etc

Arylide:
nyaadev did the tracker shit, so I suppose he should look at it.

TheAMM:
Here's a patch for the tracker api stuff:
https://gist.github.com/TheAMM/8fb6c48ceab10a3b2f085e6e9845ed52
Download, `git checkout nuke` and apply with `git am tracker_api.patch`.

CounterPillow:
@TheAMM you should be able to just push to the nuke branch on CounterPillow.

Thanks for this, so all that's left is to fix the isort (oops lol)

CounterPillow:
Restricted nuking to admins, should be ready for merge now I think.

Arylide:
Apparently nyaadev thinks it is fine after talking to him.

Issue #364, opened 2017-09-13
[Suggestion] Add pager to top of listing


hal0thane:
It's so trivial I hesitate to even mention it, but having a pager at the top of the listing would make the site easier to navigate.

When I am browsing through the listings that have been added since my last refresh, I hate getting 4 or 5 listing from the top of the screen on the 2nd or 3rd page, then having to scroll all the way down to get the link back to page 1. Having a pager ("« 1 | 2 | ... | 5 | 6 »") at the top as well as the bottom would be helpful, but it's certainly not an important request.

As far as I could tell from a quick search through the issue listing, this is not a duplicate request.

PR #391, opened 2017-10-26
Add pagination on the top of search results

Nalapl3:
Added additional pagination on the top of the search results as mentioned in issue #364. Untested, but should work. Doesn't show if there's 0 results to avoid doubling the "Displaying results 0-0 out of 0 results.".


Arylide:
Nobody seems interested in this as standard behaviour, pagination makes sense to be at the bottom, since the general use is for once you've looked through all the results a page has, you move to the next. Perhaps when user page options are fleshed out this could be a user option instead.

PR #392, opened 2017-10-28
Fix nuke button visbility

CounterPillow:
Only show it to admins, move it to its own line, and show it when users are banned.

I'll add a confirmation to the nuke button once I get around to refactoring it into its own form with its own `POST` end-point.

It's moved to its own line in preparation for future buttons like "nuke all comments".


CounterPillow:
Okay, I've done some fairly extensive refactoring.

* nuking now has its own form
* nuking now has its own end-point, protected by a new `@admin_only` endpoint
* there now is a confirmation dialog when clicking the button, which asks you to type in the case-insensitive name of the user.

I think this can be merged after a review.

CounterPillow:
Just force-pushed the last commit again to fix a CSRF vuln it introduced (oops)

Issue #326, opened 2017-08-02
Editing comments?


zeromido:
Do you have any plans to do that?

Arylide:
When someone feels like doing it I suppose. PRs welcome too.

junh1024:
I need this too.

Arylide:
Done at #396

PR #396, opened 2017-10-30
Comment editing

stgn:
For #326. Can probably be merged as-is, but there are a few things to consider:

- Only the comment poster can edit their comment. Moderators cannot edit other people's comments, and I couldn't think of any good reason they should be able to.
- Comments can be edited an unlimited number of times and after an unlimited amount of time. I'm not sure if this is desirable or not.
- It requires JavaScript. Edits are submitted via XHR, and the endpoint responds with JSON.


CounterPillow:
Thank you for your work!

>It requires JavaScript. Edits are submitted via XHR, and the endpoint responds with JSON.

That's alright, lots of other things on the site require JS too. (The dropdowns, the selects, and the markdown to be specific.)

>Comments can be edited an unlimited number of times and after an unlimited amount of time. I'm not sure if this is desirable or not.

Limiting it to maybe an hour after posting would probably be better. Editing should be used to fix typos and such. If we do get some sort of new comments feed at some point, spammers could abuse unlimited-time edits to evade our ever-watchful eyes.

TheAMM:
Late images, images are nice:
image
image
image
image

zeromido:
Many thanks! @stgn

Issue #401, opened 2017-11-04
Main page comment hover.


Avostear:
Add an effect hover effect to main search page to show some of the comments.
![untitled-1](https://user-images.githubusercontent.com/2817076/32407280-07b46fc2-c15d-11e7-9a43-eb42fa2bdd74.jpg)

TheAMM:
The idea is to keep the listing efficient on database queries, but this could be done with XHR on-request, so only when needed.
Up for someone to implement, then.

Nalapl3:
Which comments should it show? I think it should show the last ones. This would also have to much the website layout.

Avostear:
To be honest, thats got me stumped; latest, up-vote system, random, etc. I mean in a perfect world it would only show comments that actually describe the A/V but thats a stretch. Just figured it would be cool.

PR #410, opened 2017-11-10
Preview comments

Ostrea:
Show preview of comments as asked in #401. It just shows 5 earliest comments.

### DIFF
--- a/nyaa/templates/search_results.html
+++ b/nyaa/templates/search_results.html
@@ -73,7 +73,7 @@
 					{% set torrent_id = torrent.meta.id if use_elastic else torrent.id %}
 					{% set com_count = torrent.comment_count %}
 					{% if com_count %}
-					<a href="{{ url_for('torrents.view', torrent_id=torrent_id, _anchor='comments') }}" class="comments" title="{{ '{c} comment{s}'.format(c=com_count, s='s' if com_count > 1 else '') }}">
+                    <a href="{{ url_for('torrents.view', torrent_id=torrent_id, _anchor='comments') }}" class="comments">
 						<i class="fa fa-comments-o"></i>{{ com_count -}}
 					</a>
 					{% endif %}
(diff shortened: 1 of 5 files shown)


nyaa/views/torrents.py
```diff
@@ -487,3 +487,21 @@ def _get_cached_torrent_file(torrent):
             out_file.write(torrents.create_bencoded_torrent(torrent, metadata_base))
 
     return open(cached_torrent, 'rb'), os.path.getsize(cached_torrent)
+
+
+@bp.route('/comments-preview/<int:torrent_id>/', methods=['GET'])
+def comments_preview(torrent_id):
+    torrent = (models.Torrent.query
+               .options(joinedload('comments'))
```
TheAMM: While this won't kill performance, I'd much rather see it optimized to only fetch the N latest comments for a given torrent, using the query `.limit()` (instead of fetching them all and grabbing the last five).

nyaa/templates/search_results.html
```diff
@@ -73,7 +73,7 @@
 					{% set torrent_id = torrent.meta.id if use_elastic else torrent.id %}
 					{% set com_count = torrent.comment_count %}
 					{% if com_count %}
-					<a href="{{ url_for('torrents.view', torrent_id=torrent_id, _anchor='comments') }}" class="comments" title="{{ '{c} comment{s}'.format(c=com_count, s='s' if com_count > 1 else '') }}">
+                    <a href="{{ url_for('torrents.view', torrent_id=torrent_id, _anchor='comments') }}" class="comments">
```
TheAMM: Tabs!

Arylide:
Thank you, but no thank you. We aren't really trying to be a forum and there isn't a good way to do this without either harming index performance or adding additional complexity for minor gain. I should have closed the related issue earlier.

PR #417, opened 2017-11-22
Remove tracker limit and always add our trackers

TheAMM:
> With all trackers.txt trackers being included in generated .torrents, we can now be certain the magnet (which use trackers.txt) and the .torrent uses will not be split up in different swarms in case the main announce dies.
>(That is, if uploaders added enough of their own trackers and additional trackers were deemed unnecessary (at least 5 already), the magnet and .torrent would only share the main site announce)

Also `# comments in trackers.txt wow`


Arylide:
Cool and good.

PR #418, opened 2017-11-22
Alert hotmail users of the void on register page

TheAMM:
More than enough users have their verification emails lost thanks to Outlook/Hotmail shenanigans.

----

image

### DIFF
--- a/nyaa/templates/register.html
+++ b/nyaa/templates/register.html
@@ -7,7 +7,7 @@
 {% from "_formhelpers.html" import render_field %}
 
 <h1>Register</h1>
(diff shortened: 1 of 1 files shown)


CounterPillow:
Considering the amount of ESL users we have, rewording it to "Please do not use" may be easier to understand. Also if @Arylide is okay with it, making this a more visible yellow warning notification way make sense

Arylide:
Might be better to just reject hotmail,live, and outloook domains with an error, there is basically no point to even allowing it since they just block the mail.

PR #427, opened 2017-12-22
static: losslessly optimise PNG images

CounterPillow:
Used zopflipng to optimise the images, saving a whopping 8238 bytes.

The optimisation is lossless, i.e. the decoded pixel values do not change at all.


Arylide:
This is bad and you should feel bad.

CounterPillow:
(´・ω・｀)

Issue #351, opened 2017-08-30
Add a spam filtering system


CounterPillow:
This is a feature request for a spam filter functionality, and a discussion about how it should work.

Filters could take the form of regular expressions that are case insensitive. They can apply to a multitude of things:
* Torrent titles
* Torrent information fields
* Torrent descriptions
* Filenames
    - This would allow us to reject certain torrents like ones with BitComet padding files
* Comments

The resulting action could also be configurable, e.g.
* Reject torrent
* Reject torrent and IP-ban uploader
* Whitelist from other filters

If this was extended to include an additional parameter which restricts the filter to a certain category, we could do things like disallow executables in categories that shouldn't have executables.

TheAMM:
#342 works as a good foothold.
I'd prefer the rules/actions be configurable outside `config.py`, but allowing very flexible rules/actions would be a pain with "proper" tables and relationships, so I'm suggesting one table with a JSON serialization of a validator(set)/action(set) combo. Adding a nice UI would require UI work, though, but maybe we can just go with a textfield for json input or something for now.

I'm thinking of stuff like
```json
{
  "name" : "Executables in wrong category",
  "validators" : [
    {"type" : "all", "validators" : [
        { "type" : "regex", "target": "filelist", "value": "\\.exe$" },
        { "type" : "value_not", "target": "torrent_category", "value": "x_xx" }
      ]
    }
  ],
  "actions" : ["reject"]
}
```
... or something. I'm certainly over-engineering this because it's neat to do.

CounterPillow:
Okay, more armchair overengineering:

- allow filters to only apply to Anonymous uploads
- allow filters to only apply to fresh accounts (either based on time or number of torrents? Bikeshedding needed)
- action to reject with a custom message (for cases where accidentally getting caught in the spam filter is more likely)
- action to reject with a ban (for very clear cases of abuse)
- action to permit but set a cooldown on the IP (to e.g. prevent Anon floods of .exe or archives)

So, who's gonna bite the bullet and try to implement this mess?

TheAMM:
The current implementation can handle anons, fresh accounts (requires adding a User.age or allowing `eval()`s on the filters [obviously with sane envs]), rejections & bans with message.
Cooldown will require keeping track of all uploading IPs and their last date, which will reach outside the filter logic, so it's bit of a pain.

[...]

PR #440, opened 2018-01-29
Feature/shadow flagging

the-eater:
This pull request allows shadowing of torrents needed for #351 

A shadowed torrent isn't visible from the index unless you're ~~an admin~~ a moderator or the user that uploaded the torrent, torrent is still available for viewing from invidual page

TODO

* [x] ES support
* [x] UI feedback for admin roles
* [x] filter for shadowed torrents for admin roles
* [x] Add buttons for admin roles to shadow or unshadow a torrent

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ __pycache__
 
 # Editors
 /.vscode
+/.idea
 
 # Databases
 *.sql
--- a/es_mapping.yml
+++ b/es_mapping.yml
@@ -102,6 +102,8 @@ mappings:
         type: boolean
       hidden:
         type: boolean
+      shadowed:
+        type: boolean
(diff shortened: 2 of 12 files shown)


CounterPillow:
This basically just reimplements the "hidden" flag, why? Wouldn't it be better to have an "edit freeze" flag, with which hidden could be forced by moderators?

Also, the integer in models.py clashes with the comment locking PR, which uses 128 for locked status.

the-eater:
This is going to be used for the spam filter, once there is a false positive, you don't want to affect it the uploader in any real way, now the only problem with a false positive will be that the torrent won't be listed, while if you put an edit lock/freeze on it, it's very noticable and can be very annoying.

PR #444, opened 2018-02-01
Undo responsive table for reports

sfan5:
I opted for `table-responsive` since the other option was to duplicate a CSS rule already present in bootstrap dark, this PR does that instead.


Arylide:
smh.

PR #470, opened 2018-04-07
Update for dark theme

kostek00:
I was asked by Anna-Maria Meriniemi if I would like to make pull request with my [DeepDark Nyaa.si](https://userstyles.org/styles/143105/deepdark-nyaa-si) style in stylish for official Nyaa repo. So here it is. I cleaned most of it from unnesesery code. I hope it's not too chaotic and noobish.

### DIFF
--- a/nyaa/static/css/main.css
+++ b/nyaa/static/css/main.css
@@ -425,29 +425,29 @@ h6:hover .header-anchor {
 /* Dark theme */
 
 body.dark {
-    background-color: #202020;
-    color: #9A9A9A;
+	background-color: #202020;
+	color: #9A9A9A;
 }
 
 body.dark .navbar {
-    box-shadow: 0 1px 3px rgba(0, 0, 0, 0.7);
-    background-color: #191919;
+	box-shadow: 0 1px 3px rgba(0, 0, 0, 0.7);
+	background-color: #191919;
 }
 
 body.dark .navbar a,
 body.dark .navbar a:visited,
 body.dark .navbar a:focus{
-    color: #9A9A9A;
+	color: #9A9A9A;
 }
 
 body.dark .navbar a:hover {
-    color: #ffffff;
+	color: #ffffff;
 }
 
 body.dark kbd,
 body.dark .btn.edit-comment {
-    background-color: #4a4a4a;
-    color: #ffffff;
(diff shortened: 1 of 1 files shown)


TheAMM:
Nice! I'll give this a better look later. Some of the rows will require colors (hidden and deleted torrents which the normal users don't really get to see) and IMO some of the `:visited` colors are too dark to separate from the background. But I consider this an improvement on the current dark CSS.

Here are some screenshots:
image
image
image

kostek00:
I was wondering if hidden torrents are hidden even for admin (I was assuming they aren't but still). When it comes to deleted torrents I see only 2 ways. Leave them as bright grey with dark font color or give them some different color. When it come to `:visited` I was thinking to make them the same color as `not :visited` but I decided to give them darker color. I will see tommorow what can I do with that but it will be hard.

kostek00:
I forgot to ask. Does hidden torrents have it's own class on torrent list or only in line style as in torrent view? Because I do have color theme for it already implemented in my style. I just don't know what I have to target.

TheAMM:
Here's the template line for the torrent CSS class:
https://github.com/nyaadevs/nyaa/blob/master/nyaa/templates/search_results.html#L60
It is a bit hacky, since we're using the bootstrap defaults but also provide our own `deleted`.

Here's the torrent page's class choosing:
https://github.com/nyaadevs/nyaa/blob/master/nyaa/templates/view.html#L9
Right now we use an inline style for deleted torrents on the torrent page, so don't worry about that.

kostek00:
I see. Everything is clear now. I will make those today or tommorow.

kostek00:
I will explain changes.

Instead of making `:visited` links darker I took different approach. I make them almost greyed out. They do retain a little bit of their color but I doubt that anyone will see difference. I couldn't make it any other way otherwise links would be too dark. There is another way to deal with them. Instead of changing their color they can have checkmark before link with image. Most browsers won't allow for `:visited:before` so there is no point in doing so with `content`.

For deleted torrents I didn't use another pair of grey colors. Belive me you don't want grey rainbow you can go crazy looking at that. On the other hand I didn't want to use light grey to make bigger difference between them and not deleted because it's not how dark theme should look like. Instead I make them blue.

PR #480, opened 2018-04-18
add Deleted, Trusted and Remake info to RSS feed

nallown:


CounterPillow:
If it's deleted, it shouldn't show up in the RSS at all, shouldn't it?

nallown:
If it's deleted, then it shows as deleted on the homepage, I'm just using the code nyaa already uses on the search results.

The snippet below is from the nyaa search page (basically the homepage).

https://github.com/nyaadevs/nyaa/blob/0cc25c3569d22b2734730682aa14ad872aa3a6e7/nyaa/templates/search_results.html#L60

nallown:
Closing this pull request, I have a better solution

PR #481, opened 2018-04-18
add trusted and remake information to RSS feed

nallown:


Manjimon:
This would be nice, yeah. Accept it thanks~

disturbed01:
this looks pretty sweet, will simplify my life. please accept this

nyaa/templates/rss.xml
```diff
@@ -40,6 +40,11 @@
 			<nyaa:categoryId>{{- cat_id }}</nyaa:categoryId>
 			<nyaa:category>  {{- category_name(cat_id) }}</nyaa:category>
 			<nyaa:size>      {{- torrent.filesize | filesizeformat(True) }}</nyaa:size>
+			<nyaa:submitter> {{- 'Anonymous' if torrent.anonymous or not torrent.user else torrent.user.username }}</nyaa:submitter>
+			<nyaa:trusted>	 {{- torrent.trusted }}</nyaa:trusted>
+			<nyaa:remake>	 {{- torrent.remake }}</nyaa:remake>
+			<nyaa:summary><![CDATA[{{- torrent.description or '#### No description.' }}]]></nyaa:summary>
```
TheAMM: Not available on [ES results](https://github.com/nyaadevs/nyaa/blob/master/es_mapping.yml#L84).

nyaa/templates/rss.xml
```diff
@@ -40,6 +40,11 @@
 			<nyaa:categoryId>{{- cat_id }}</nyaa:categoryId>
 			<nyaa:category>  {{- category_name(cat_id) }}</nyaa:category>
 			<nyaa:size>      {{- torrent.filesize | filesizeformat(True) }}</nyaa:size>
+			<nyaa:submitter> {{- 'Anonymous' if torrent.anonymous or not torrent.user else torrent.user.username }}</nyaa:submitter>
```
TheAMM: `torrent.user` won't work on ES - you'll have to work around it as above with the category id.

nyaa/templates/rss.xml
```diff
@@ -40,6 +40,11 @@
...
+			<nyaa:trusted>	 {{- torrent.trusted }}</nyaa:trusted>
+			<nyaa:remake>	 {{- torrent.remake }}</nyaa:remake>
```
TheAMM: These should rather be explicit strings, and instead of `True` and `False` I'd prefer `Yes` and `No` (`torrent.remake and 'Yes' or 'No'`).

nyaa/templates/rss.xml
nallown: How can I make available on Elastic Search?

nyaa/templates/rss.xml
nallown: I checked the code, on Elastic Search you use `torrent.sub_category_id` instead of `torrent.sub_category.id_as_string`. How did you make the `sub_category.id_as_string` value available in `torrent.sub_category_id`?

I'm guessing I need to do something similar with the torrents username. Something like:
```j2
{% set submitter = use_elastic and torrent.submitter or torrent.user.username %}
```

I'm not sure how I can map the username to `torrent.submitter` though

nyaa/templates/rss.xml
nallown: Fixed.

nyaa/templates/rss.xml
TheAMM: You can't. We don't store the description in ES, and it's not really worth the overhead to add it just for RSS.

nyaa/templates/rss.xml
TheAMM: We don't make anything look like anything. You can see [just above](https://github.com/nyaadevs/nyaa/pull/481/files/29ccd939f8ac0f906804d57ea41e3607c5795397#diff-0c21456d97f5b069933b89256c6c1573R39) how we choose either the ES category field or the DB model's relationship field. You don't have to map that back to anything, just use the variable you just `set`.

nyaa/templates/rss.xml
nallown: Should I make descriptions only available in none ES queries?

nyaa/templates/rss.xml
nallown: I get that but like how's the fields made? Or where can I find the ES field for username?

nallown: I checked the ES Mappings and I saw you have a `uploader_id` in there. Is that the user's username or his account ID?

nyaa/templates/rss.xml
TheAMM: Nah. It breaks consistency, pure-db queries are rate for RSS, few would have any use for the field (it's not visible for human-RSS readers), plus it's unprocessed markdown.

nallown: You're right people don't really need the summary since they can use the other info from the feed. I'll remove it then since I don't need it either.

What should I do about your other change request? The one for the submitter field?

nyaa/templates/rss.xml
TheAMM: `torrent` is either `models.Torrent` or an ES object with the fields listed in https://github.com/nyaadevs/nyaa/blob/master/es_mapping.yml#L84 when `use_elastic` is `True`.

nallown: I already looked at the ES mappings but found no field for username or submitter, what value am I meant to show on ES?

TheAMM: Oh shit, you're right, I remembered falsely (since `uploader_id` is there). This will complicate things, since we'll have to decide whether or not to store the uploader name in ES. I'll get back on this later.

nallown: Any updates? If its too much work then I can remove the "submitter" from the pull requests since its not really needed. If someone needs it then it can be added later in another pull requests.

nallown: @TheAMM Can you give me update

nallown:
@TheAMM I've changed the pull request and removed the "submitted" and "description". I've also made every other change that you asked me to make with the rest.

PR #483, opened 2018-04-23
Update bootstrap-dark.min.css

Exertt:
Dark mode - changed link color from blue to white; changed link focus and hover color to slightly shadow from dark blue.

This slight change of colours makes it much more readable for people using dark mode.


TheAMM:
Pics!

PR #494, opened 2018-05-16
Update leechers-paradise tracker domain

Nalapl3:
Leechers-paradise.org changed their primary domain. [Their website](http://leechers-paradise.org/) says "I also moved the primary domain from leechers-paradise.org to open-internet.nl." It also says old domains will work, but perhaps it is still better to change it.


Arylide:
It says on there that the old domain will continue to work, and considering it has existed for all this time anyway, I don't see any need to change it right this instant. It will be changed when the new domain has been adopted more.

Nalapl3:
Keeping the old domain isn't exactly helpful for adopting the new one.

PR #503, opened 2018-06-13
Add 1 hour cache to magnet URIs

CounterPillow:
Using Flask-Caching, we can memoize the `magnet_uri` method. Here, a timeout of 1 hour is chosen, though that value can be fiddled with.

The cache is defined in `extensions.py`, but gets initialised in `__init__.py`.


Arylide:
How does github work

PR #507, opened 2018-07-09
[Schema change] Add registration IP

CounterPillow:
This will keep track of which IP a user has registered from, to prevent evading rangebans. It will only be shown to admins.


rmrhz:
Would there be IPv6 support?

rmrhz:
@CounterPillow woah. i never knew. i'm getting outdated with lengths.

PR #518, opened 2018-08-19
[Config change] Allow blacklisting email servers by IP

TheAMM:
```
Adds a new config entry (EMAIL_SERVER_BLACKLIST, tuple of IPv4 addresses
as strings) and an email validator for registering, which will query all
the MX records for the domain and reject the registration if any of the
A records for any of the MX records are found in the blacklist.
If the query fails, the blacklist is ignored; the email is accepted.
```

This also adds `dnspython` as a requirement; run `pip install -r requirements.txt`


Arylide:
Cool and good.

Issue #519, opened 2018-08-24
Login is vulnerable to the open redirect vulnerability


NyanKiyoshi:
Hello,

I discovered in the code that there is no check on the user submitted redirect URL for the login view. Which makes your project vulnerable to the open redirect vulnerability.

Here is the faulty code:
https://github.com/nyaadevs/nyaa/blob/8c892f09ccdf168b210a0ec861598bec43cdc549/nyaa/views/account.py#L216-L218

Here is how to reproduce it:
1. Logout;
1. Go to https://nyaa.si/login?next=https://example.com;
1. Enter your credentials and hit 'Login';
1. You should be redirected to `https://example.com` instead of having an error or being redirect to the index page.

There are multiple ways to prevent this from happening. One of them could be doing this: http://flask.pocoo.org/snippets/62/, but this solution remains not perfect.

PR #523, opened 2018-09-16
Tiny patch to prevent open redirect vunerability.

ojiii:
Fixed the issue mentioned in #519 by adding a check that only allows redirects for URL with host name, if ```?next=``` URL does not contain host, it will redirect to ```main.home```.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 __pycache__
 /nyaa/static/.webassets-cache
 
+# Byte Code
+*.pyc
+
 # Virtual Environments
 /venv
 
--- a/nyaa/views/account.py
+++ b/nyaa/views/account.py
@@ -217,7 +217,7 @@ def redirect_url():
         flask.request.referrer or \
         home_url
 
-    if url == flask.request.url or not flask.request.host in url:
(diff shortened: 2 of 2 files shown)


TheAMM:
1. Python 3 stores bytecode in `__pycache__`, which is already ignored. The `*.pyc` is redundant.
2. This check can easily be circumvented by adding the host in the malicious redirect (ala `https://fake.com/real.com/asd`), so it doesn't really help.

ojiii:
@TheAMM I will properly handle the patch to prevent the open redirect vunerability.

Regarding the *.pyc, when I ran the project it created a bunch of __init__.pyc that is why I added it to prevent it from getting pushed into the repository.

TheAMM:
Already done in b2ddba994ca5e78fa5dcbc0e00d6171a44b0b338.
It sounds like you were using Python 2 - this project uses (at least) 3.6. You can also add files individually to the commit, which I consider handy for that sort of self-review.

ojiii:
I used Python 3 because I read the documentation. I realized what was wrong, I used pyvenv instead of pyenv, for some reason it generates *.pyc files.

PR #527, opened 2018-10-21
backend: blacklist Windows reserved filenames

CounterPillow:
Windows has a few special filenames that it does not allow the explorer.exe and command line to see, but can still be created by applications. This is due to some jank DOS compatibility.

These filenames can be abused to troll Windows users, so we should probably blacklist them.

### DIFF
--- a/nyaa/backend.py
+++ b/nyaa/backend.py
@@ -14,6 +14,21 @@ from nyaa.extensions import db
 
 app = flask.current_app
 
+# Blacklists for _validate_torrent_filenames
+# TODO: consider moving to config.py?
+CHARACTER_BLACKLIST = [
+    '\u202E',  # RIGHT-TO-LEFT OVERRIDE
+]
+FILENAME_BLACKLIST = [
+    # Windows reserved filenames
+    'con',
+    'nul',
+    'prn',
+    'aux',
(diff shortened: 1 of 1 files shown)


nyaa/backend.py
```diff
@@ -62,16 +77,14 @@ def _recursive_dict_iterator(source):
...
-    ]
+    ''' Checks path parts of a torrent's filetree against blacklisted characters
+        and filenames, returning False on rejection '''
     file_tree = json.loads(torrent.filelist.filelist_blob.decode('utf-8'))
 
     for path_part, value in _recursive_dict_iterator(file_tree):
-        if any(True for c in character_blacklist if c in path_part):
+        if path_part.rsplit('.', 1)[0].lower() in FILENAME_BLACKLIST:
+            return False
+        if any(True for c in CHARACTER_BLACKLIST if c in path_part):
```
FichteFoll: Time to try this new suggestion feature.

```suggestion
        if any(c in path_path for c in CHARACTER_BLACKLIST):
```

CounterPillow: It's `path_part`, not `path_path`. This code also has nothing to do with the actual proposed change in this PR, so I don't know what you're doing here.

FichteFoll: I was just making a suggestion with the mentioned new feature from Github. Fixed the typo.

Issue #496, opened 2018-05-31
Feature request: comment count in RSS feed


rr-:
While we still haven't any sort of notifications, with this feature, we could code one ourselves without resorting to parsing HTML which is grey area.

PR #528, opened 2018-10-25
Add amount of comments to RSS

Nalapl3:
Implements #496.

### DIFF
--- a/nyaa/templates/rss.xml
+++ b/nyaa/templates/rss.xml
@@ -40,7 +40,7 @@
 			<nyaa:categoryId>{{- cat_id }}</nyaa:categoryId>
 			<nyaa:category>  {{- category_name(cat_id) }}</nyaa:category>
 			<nyaa:size>      {{- torrent.filesize | filesizeformat(True) }}</nyaa:size>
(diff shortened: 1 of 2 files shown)


nyaa/templates/rss.xml
```diff
@@ -40,6 +40,7 @@
 			<nyaa:categoryId>{{- cat_id }}</nyaa:categoryId>
 			<nyaa:category>  {{- category_name(cat_id) }}</nyaa:category>
 			<nyaa:size>      {{- torrent.filesize | filesizeformat(True) }}</nyaa:size>
+			<nyaa:comments>      {{- torrent.comments | count }}</nyaa:comments>
```
TheAMM: This will be horribly horribly inefficient, as it'll prompt SQA to lazy-load all the comments and wrap them in classes and put them in a list just for Jinja to count them. Even an optimized query to count the comments beforehand (every request!) will bring an undesirable performance impact for a niche feature.
However, luckily, we *just* might have [something designed for this](https://github.com/nyaadevs/nyaa/blob/master/nyaa/models.py#L129).

PR #531, opened 2018-10-30
Tweaks for "Complete" torrents

Nalapl3:
Currently, "Complete" option has basically no effect. You can search for complete only, but this option is hidden and unknown.

This PR adds "Complete only" to dropdown + a marker for torrents marked as complete. I don't have any better idea where to put that marker, but I think a separate row is the best. It also allows for later addition of more fields, which can be used with trusted/remake eg. trusted+remaster.

[Here](https://cdn.discordapp.com/attachments/191116708990287872/506935143529971740/unknown.png)'s an example of how it looks.

Even if you decide to not merge this layout change with a new row, I believe it's worth adding a "Complete only" option to dropdown to encourage people to use it.


Nalapl3:
I have also tried putting this [C] icon into the name field, like some other websites do, but it didn't look very good IMO.

CounterPillow:
Okay, so I'm closing this because we're unlikely to benefit anything from the "Completed" flag. Basically, it was a mistake to add it, and it's not likely to be of any use even with this PR, reason being that 99% of past content isn't flagged properly and never will be. So there's not much point to it in terms of search. The colour in the result listing also conflicts with what mods see hidden torrents as, and will just be abused by idiots to stand out from the masses instead of being used properly.

PR #533, opened 2018-11-21
Add trusted application functionality

CounterPillow:
This lets users apply for trusted status, given certain minimum requirements. Moderators can then review the applications, giving a recommendation, and administrators can accept or reject them.

If an application is accepted or rejected, the user receives an e-mail about it.

Markdown images are not rendered in applications to prevent browsers from sending automatic requests to untrusted webservers.

Users who have had their application rejected cannot re-apply for a set amount of days.

### DIFF
--- a/nyaa/models.py
+++ b/nyaa/models.py
@@ -650,8 +650,8 @@ class User(db.Model):
         for ts_flavor, t_flavor in ((NyaaStatistic, NyaaTorrent),
                                     (SukebeiStatistic, SukebeiTorrent)):
             uploads = db.session.query(func.count(t_flavor.id)).\
-                    filter(t_flavor.user == self).\
-                    filter(t_flavor.flags.op('&')(int(TorrentFlags.REMAKE)).is_(False)).scalar()
+                filter(t_flavor.user == self).\
+                filter(t_flavor.flags.op('&')(int(TorrentFlags.REMAKE)).is_(False)).scalar()
             dls = db.session.query(func.sum(ts_flavor.download_count)).\
                 join(t_flavor).\
                 filter(t_flavor.user == self).\
--- a/nyaa/models.py
+++ b/nyaa/models.py
@@ -11,8 +11,9 @@ from urllib.parse import urlencode
 import flask
 from markupsafe import escape as escape_markup
 
-from sqlalchemy import ForeignKeyConstraint, Index
+from sqlalchemy import ForeignKeyConstraint, Index, func
 from sqlalchemy.ext import declarative
+from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy_fulltext import FullText
 from sqlalchemy_utils import ChoiceType, EmailType, PasswordType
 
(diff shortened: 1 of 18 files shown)


nyaa/models.py
```diff
@@ -847,6 +867,77 @@ def is_rangebanned(cls, ip):
...
+class TrustedApplicationStatus(IntEnum):
+    # If you change these, don't forget to change is_closed in TrustedApplication
+    NEW = 0
+    REVIEWED = 1
+    ACCEPTED = 2
+    REJECTED = 3
+
+
+class TrustedApplication(db.Model):
+    __tablename__ = 'trustedapps'
```
TheAMM: `__tablename__ = 'trusted_applications'`

nyaa/views/account.py
```diff
@@ -223,6 +223,48 @@ def profile():
...
+    if not flask.g.user:
+        return flask.redirect(flask.url_for('account.login'))
+    trusted_form = None
+    deny_reasons = []
+    if flask.request.method == 'POST':
+        trusted_form = forms.TrustedForm(flask.request.form)
+        if trusted_form.validate():
+            ta = models.TrustedApplication()
+            ta.submitter_id = flask.g.user.id
+            ta.why_want = trusted_form.why_want_trusted.data
```
TheAMM: I'd `.strip()` these two just because, although I forget if WTF passes the minimum length even if it's spaces.

nyaa/views/account.py
```diff
@@ -223,6 +223,48 @@ def profile():
...
+            db.session.add(ta)
+            db.session.commit()
+            flask.flash('Your trusted application has been submitted.', 'success')
+            return flask.redirect('/trusted')
+    else:
+        if flask.g.user.is_trusted:
+            deny_reasons.append('You already are trusted.')
+        if not flask.g.user.satisfies_trusted_reqs:
+            deny_reasons.append('You do not satisfy the minimum requirements.')
+        if (models.TrustedApplication.query.
```
TheAMM: `.first()` is fine instead of `.count() > 0`, although if you chose this for performance, :woman_shrugging:.

nyaa/views/account.py
CounterPillow: Do we get any issue if we don't strip them? In the very synthetic case of stripping them, it could break code blocks at the start of the field opened with 4 spaces. Obviously not something to worry about in this context though.

nyaa/models.py
```diff
@@ -643,6 +644,25 @@ def created_utc_timestamp(self):
...
+            uploads = db.session.query(func.count(t_flavor.id)).\
+                    filter(t_flavor.user == self).\
+                    filter(t_flavor.flags.op('&')(int(TorrentFlags.REMAKE)).is_(False)).scalar()
+            dls = db.session.query(func.sum(ts_flavor.download_count)).\
+                join(t_flavor).\
+                filter(t_flavor.user == self).\
+                filter(t_flavor.flags.op('&')(int(TorrentFlags.REMAKE)).is_(False)).scalar()
+            num_total += uploads or 0
+            downloads_total += dls or 0
+        print("Uploads: {}, DLs: {}".format(num_total, downloads_total))
```
CounterPillow: oops I should probably get rid of this

CounterPillow:
Do you think I should index `submitter_id` in `TrustedApplication`? We do filter by it quite a lot, though I do not know if this is relevant enough.

nyaa/views/account.py
```diff
@@ -223,6 +223,48 @@ def profile():
...
 
+@bp.route('/trusted/request', methods=['GET', 'POST'])
+def request_trusted():
+    if not flask.g.user:
+        return flask.redirect(flask.url_for('account.login'))
+    trusted_form = None
+    deny_reasons = []
+    if flask.request.method == 'POST':
+        trusted_form = forms.TrustedForm(flask.request.form)
+        if trusted_form.validate():
```
Hamuko: `deny_reasons` is never computed/checked for POST requests, so if you don't pass minimum requirements, you can just directly POST the data.

CounterPillow: They would have a hard time getting past trusted_form.validate() without a csrf token which they never get because they never get the form, but sure, I've changed to code so that it checks the `deny_reasons` both paths.

nyaa/views/account.py
TheAMM: `r.strip()`, then. It's simply a small useless abuse vector, pasting 4k spaces. 4k of data is not much to worry about, but hey, it's user input, and it's clean to `.rstrip()` it.

CounterPillow:
I've addressed AMM's comments and rebased the migration on top of @nyaadev's. If someone reviews this and finds it to be satisfactory, then I think this is ready to merge.

Arylide:
Wow

PR #535, opened 2018-12-06
Add Vagrant support

LuxF3rre:
Added Vagrant support for creating the uniform development environment. Added a more aggressive gitignore rule for python cache. I don't know if it is directly connected to how Vagrant works.

The provisioning script is following the instructions included in readme.

Elasticsearch support is yet to be added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,14 @@
 # Cache
 __pycache__
 /nyaa/static/.webassets-cache
+/nyaa/*.pyc
 
 # Virtual Environments
 /venv
+
+# Vagrant
+/.vagrant
+ubuntu-bionic-18.04-cloudimg-console.log
 
 # Coverage
 .coverage
--- a/README.md
+++ b/README.md
@@ -1,6 +1,23 @@
 # NyaaV2 [![Build Status](https://travis-ci.org/nyaadevs/nyaa.svg?branch=master)](https://travis-ci.org/nyaadevs/nyaa)
 
 ## Setting up for development
+### Vagrant
+
+The easiest way to start developing is by using Vagrant. Install Vagrant and Virtualbox, and then `cd` into the project directory and run `vagrant up`. This will create and configure a virtual machine. To start the server run:
+
+1. `vagrant ssh` - connect to the virtual machine
+
+2. `su` and enter a blank password for root - login as root
+3. `cd /vagrant` - change directory to the project directory
+4. `pyenv activate nyaa` - activate the virtual environment for python
+5. `python run.py` - run the server
+
+When you finish developing, simply Control + C to stop the server, then `exit` to close ssh connection and then `vagrant halt` to shutdown the virtual machine. To start it again, `cd` into the project directory and run `vagrant up`. 
+
+Note: currently elasticsearch is not supported out of the box.
+
+### Manual
+
 This project uses Python 3.6. There are features used that do not exist in 3.5, so make sure to use Python 3.6.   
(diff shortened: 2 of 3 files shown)


CounterPillow:
Does this actually make anything easier? To me it just looks like it puts more burden on us to keep maintaining the vagrant file, and makes development more complicated by recommending to do it in some nonsense virtual machine.

If someone who actually wants to contribute and will contribute can say "yes this makes my life better" then sure, but otherwise I'm sceptical of this.

such-doge:
This is not necessary for our needs, and no one on or off the team has expressed to us that automatic dev environment management would be desirable. I'm closing this.

LuxF3rre:
I see your points and it is up to you to accept the pull request but consider the following:
* Using Vagrant is an option. You can still use your own environments or create it on your machine and in that case you won't see any difference. Yet for people working on Windows or Mac it would make things much easier - no need to write "setting for development" part only for Windows and Mac users. Also, this solution addresses the issue that vagrant solves: platform's specific design and problems - easy replication of bugs - everyone has the very same environment be it Windows, Mac or Linux. It is up to you to decide what is your contribution policy (how much you make easier for everyone else to contribute) or do you want it in the first place. If you want to keep the project mainly maintained by you with little to no contribution from others, you are right, there is no point in using Vagrant. Otherwise, I see two possible solutions: postpone using Vagrant (or any other similar solution) and wait for the community to grow in order to make it useful enough to keep it maintained; start using it now.
* I understand that such a feature requires more work from you to keep it up-to-date and working. I reckon that doing so will not be very troublesome. This Vagrantfile uses a shell script as a provisioner. If you were to change a development environment (let's say change database to PostgreSQL), you mostly would need to paste commands that you would have to include in the readme to Vagrantfile. Moreover, it would be easier for everyone else to switch (just `vagrant up` and you are good to go) and you won't clutter your system with programs you no longer use as those environments are separated. And lastly, stopping using vagrant requires deletion of Vagrantfile and some lines from .gitignore and readme. Not much would change as you would also add manual installation instructions. 
* I thought about using Vagrant as a first step to adding more tools like yarn, postcss or gulp to streamline the optimization and workflow - automatic creation of sprites for deployment while still using regular images in development etc. I found it too cumbersome require others to install those tools and perhaps some other as well on their machines just for one project or because they want to make a small contribution.

These are my reasons. Thank you for considering those changes.

CounterPillow:
My main issue is that people will think of Vagrant as the supported way of hacking on Nyaa, and will ask for support with issues pertaining to doing Nyaa development within the context of Vagrant. None of the main contributors will be using Vagrant, and the Vagrant file was added by a one-time contributor.

Drive-by contributors love submitting PRs that make a project use their deployment management system du jour, be it docker, vagrant or ansible. It doesn't add as much value as you think it does. It mostly just adds another layer of abstraction you inevitably will have to fight with when it doesn't work the way you want it to work.

If a person is too lazy to spend the 60 seconds it takes to set up a Nyaa dev environment, then will they really make a good contributor? 

>I thought about using Vagrant as a first step to adding more tools like yarn, postcss or gulp to streamline the optimization and workflow

We already have an automatic system for minimising JS and potentially CSS with flask-assets or whatever it was called, adding another one wouldn't make much sense.

>automatic creation of sprites for deployment

not needed, we don't host our own icons (on purpose). Also no longer needed with HTTP/2.0 AFAIU.

Feel free to use Vagrant for your own development on Nyaa. But I don't see why Nyaa contributors should maintain a vagrant file when no contributor uses vagrant.

PR #536, opened 2018-12-09
Remove leechers-paradise from default trackers

menma1234:
Leechers Paradise shut down yesterday: https://torrentfreak.com/huge-torrent-tracker-calls-it-quits-after-12-years-citing-article-13-181207/

No point continuing to add it to torrents.


menma1234:
@Arylide Sorry for the late response, but done.

Issue #541, opened 2019-01-12
Special characters break RSS feed


kevin-stuart:
Today I noticed that the item https://nyaa.si/view/1109229 seems to be breaking the RSS feed. The title contains the character 0x1D, which causes parsers such as browsers or other XML parsers to reject the XML. They report it as "PCDATA invalid Char value". So it seems that it would be reasonable to either strip or escape characters that are outside of the valid ranges when preparing the RSS feed from entries. Otherwise, bad submissions could break automated tools that expect valid feeds in a strict manner.

ReimuHakurei:
I will look into this issue and will report back in a few hours.

ReimuHakurei:
According to the W3C's RSS feed validator, RSS feeds including that item appear to be valid RSS as far as I can tell.

Can you link an example of an RSS feed that is invalid?

Additionally, here is a link to the W3C's feed validator:

https://validator.w3.org/feed/

If there is in fact a bug in our code and I have missed it, it should be fixed. However, we are not interested in working around bugs in third-party clients due to said clients not complying with standards.

kevin-stuart:
It seems the invalid character is now gone from the title of the item I linked. Of course, if the item is corrected or gone from the feed because it got pushed out, it will not trigger the problem, hence I cannot test it. But neither Chrome, nor the Java XML parser would accept the feed 6 hours ago when the problem existed. I do not think that those are minor clients.

I think the problem is that it's not an RSS problem, but an XML problem. According to the error message, it is not valid to have certain characters like this one in PCDATA sections. We could generate an XML document and insert that character by hand I see what parsers accept it. If you look at the [XML 1.1 specification in section 2.2](https://www.w3.org/TR/2006/REC-xml11-20060816/#charsets) it seems that the valid ranges are given there. So a parser is correct if it is strict and rejects documents that have characters that are not in the valid range.

ReimuHakurei:
Alright, that does in fact seem to be a bug. We'll look into fixing it.

kevin-stuart:
It seems to be an edge case that almost never occurs. I guess the bad part is that those bad characters get into the title. That could cause problems elsewhere.

I didn't want to be right about the correctness, I would not have reported it if I was not sure that it was not handled correctly. I only wanted to report a problem I observed. I could work around it by removing the bad characters before I pass them to a parser, but I think a lot of correct RSS implementation could break if an entry with a malicious title it posted. And that is something we can prevent.

PR #543, opened 2019-01-13
Fix: Special characters break RSS feed #541

ReimuHakurei:


nyaa/views/main.py
```diff
@@ -204,6 +212,10 @@ def home(rss):
 
 
 def render_rss(label, query, use_elastic, magnet_links=False):
+    # Strip invalid XML characters from title.
+    for item in query:
+        item.display_name = escape_xml_illegal_chars(item.display_name)
```
TheAMM: This is scary, you sure you want to do this instead of escaping the title in the template?

ReimuHakurei: i mean it works fine here

TheAMM: Yes, but you're mutating the database models (the ES wrapper objects are whatever), which is scary because they're in the session and all. I'd rather just use `escape_xml_illegal_chars` in the RSS template, so nothing is being mutated.

TheAMM:
We could also consider sanitizing the title on upload/edit, if these characters are bogus (I don't know if they are).

ReimuHakurei:
Some but not all of them are. Some of them are otherwise valid.

nyaa/views/main.py
CounterPillow: +1 with AMM here, this definitely belongs in the template.

TheAMM:
Alternate solution in #545

PR #544, opened 2019-01-14
Feature: docker development environment

aheadley:
Adds a Dockerfile and docker compose file to make setting up a development environment easier. Entirely self-contained in the `.docker` dir and includes some documentation in the `.docker/README.md` file.


CounterPillow:
The config file changes often, adding new keys and such. It might be better to copy the sample config file from the root of the repo instead of duplicating it again.

CounterPillow:
Also we don't use chihaya in production anymore so that's kind of a moot one. I don't know if the custom-written tracker we use is published anywhere (cc @nyaadev )

aheadley:
Ok, I've removed the references to Chihaya, and changed the config file to be generated by copying `config.example.py` and then just appending the needed values.

Also improved the start up of the sync script so it doesn't fail like 50 times before ES and mysql finish starting up and removed the test users since it didn't actually work.

CounterPillow:
Okay so basically my only concern that remains is: are you going to keep this thing working like a year down the road? Because if we do merge it and it's not maintained at all then there's no point in even having it in-tree.

aheadley:
Copy/paste of IRC discussion (modulo irrelevant messages):

~~~~
2019-01-15T11:36:39-0500 <+CounterPillow> Why does my mom insist on coming into my room and touching me while I'm trying to focus >:( >:( I'M A SPERG I CAN'T DEAL WITH THIS
2019-01-15T11:46:06-0500 < aheadley> CounterPillow: re: PR response
2019-01-15T11:46:33-0500 <+CounterPillow> write it in the github issue
2019-01-15T11:46:36-0500 <+CounterPillow> err, PR
2019-01-15T11:46:45-0500 <+CounterPillow> because if I merge this just like this it'll look weird
2019-01-15T11:47:08-0500 <+CounterPillow> and then I get spanked by mommy Aureolin and she has very strong arms
2019-01-15T11:47:45-0500 < aheadley> I could say yes but it would be meaningless, and i would argue that it is wrong to ask that question
2019-01-15T11:48:25-0500 <+CounterPillow> aheadley: imho it's also wrong to assume we'd maintain an ansible playbook, a vagrant file and a docker file, while never having used them
2019-01-15T11:48:35-0500 < aheadley> i understand where you're coming from but given that it's entirely self-contained, removing it if it is unmaintained is trivial
2019-01-15T11:49:21-0500 < aheadley> you wouldn't ask this of other contributions and this may encourage other developers to help out
2019-01-15T11:50:26-0500 <+CounterPillow> in my experience docker makes everything harder
~~~~

In summary:
  - That's a pointless question to ask because who knows what is going to happen in a year, an answer now is completely non-binding
  - The Docker stuff is entirely self-contained and trivial to remove if unmaintained
  - It may encourage new developers to join the project
  - It would be ridiculous to ask this of other contributions

CounterPillow:
Squashed and merged, for the following reasons:

1. it's self-contained and can easily be thrown out, and isn't recommended as the one true way to dev on nyaa.
2. it stops other devops shittery du jour invading the repo if we can just point towards docker and say "use that".

PR #545, opened 2019-01-15
upload/edit: sanitize some string fields

### LINKED ISSUE #541: Special characters break RSS feed

TheAMM:
This commit introduces a regex to replace illegal (expectedly unused)
characters from torrent display name, information link and description
upon upload or edit.

Fixes #541

No automatic cleanup for existing possibly dirty fields, though that's easy enough to do by hand.

### DIFF
--- a/nyaa/backend.py
+++ b/nyaa/backend.py
@@ -1,5 +1,6 @@
 import json
 import os
+import re
 from datetime import datetime, timedelta
 from ipaddress import ip_address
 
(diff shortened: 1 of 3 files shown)


nyaa/backend.py
```diff
@@ -29,6 +30,14 @@
     'lpt0', 'lpt1', 'lpt2', 'lpt3', 'lpt4', 'lpt5', 'lpt6', 'lpt7', 'lpt8', 'lpt9',
 ]
 
+# Invalid RSS characters regex, used to sanitize some strings
+ILLEGAL_XML_CHARS_RE = re.compile(u'[\x00-\x08\x0b\x0c\x0e-\x1F\uD800-\uDFFF\uFFFE\uFFFF]')
+
+
+def sanitize_string(string, replacement='?'):
```
CounterPillow: Could we use `\uFFFD` a.k.a. � here? It may be clearer to see "F��" than to see "F??", so if it's allowed in the XML spec I'd be for it.

TheAMM: It's outside the regex, at least.

PR #547, opened 2019-02-03
Fix category icon display issue.

ReimuHakurei:


CounterPillow:
Confirmed to be working, merged. Not sure if the use of inline style is the optimal solution but it's probably fine.

PR #550, opened 2019-02-23
account: force ASCII usernames on login form

CounterPillow:
Our database doesn't like it when we check for unicode data in a column that stores ASCII data, so let's stop it before it gets that far.

### DIFF
--- a/nyaa/views/account.py
+++ b/nyaa/views/account.py
@@ -26,6 +26,9 @@ def login():
             return flask.redirect(flask.url_for('account.login'))
 
         username = form.username.data.strip()
+        if not username.isascii():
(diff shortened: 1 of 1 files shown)


CounterPillow:
Apparently this needs Python 3.7

kayabe:
you can also validate it using a predefined list of characters

CounterPillow:
Yes, but I'd rather not do that, since it's more verbose and slower. Instead, I'm currently working on moving everything to python 3.7 and testing it.

CounterPillow:
Closing this since it was merged in https://github.com/nyaadevs/nyaa/commit/a96c6e84c3739faece086dbee48beef3b41b11ad

PR #567, opened 2019-05-04
Untangle users

### LINKED ISSUE #385: Untangle views/users.py
Currently, users.py is somewhat tangled up, because the ban form POSTS to the page itself in order to do actions, and all the permission checking is done in the view depending on the actions. This could be prone to breakage.

Ideally, we have a separate endpoint for each action (e.g. `/user/foo/ban`,  `/user/foo/nuke`) which then does the permissions checks as the very first step, e.g. inside a decorator. This would simplify the codepath for the view, making it easier to maintain and verify that it's correct, and stop the awkward side-effect of a refresh asking you whether you want to resubmit a form action after you ban somebody.

Opening this issue so I (and others) don't forget about this on a rainy day when some refactoring is in order.

ChrisWeaver1:
resolve #385 

- Creates endpoints for ban, unban, change user levels & removes them from view_user.
- Adds decorator for moderator+ levels

### DIFF
--- a/nyaa/forms.py
+++ b/nyaa/forms.py
@@ -453,6 +453,7 @@ class UploadForm(FlaskForm):
 
 class UserForm(FlaskForm):
     user_class = SelectField('Change User Class')
+    sumbit_user_class = SubmitField('Apply')
     activate_user = SubmitField('Activate User')
 
     def validate_user_class(form, field):
--- a/nyaa/templates/user.html
+++ b/nyaa/templates/user.html
@@ -48,7 +48,34 @@
 				{{ admin_form.csrf_token }}
 				<div class="row">
 					<div class="col-md-6">
-						{{ render_menu_with_button(admin_form.user_class) }}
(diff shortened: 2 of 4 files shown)


ChrisWeaver1:
Not sure what to do about that lint error, ipbanned is used further down the function in a render_template.
lint

CounterPillow:
Aside from some suspected permissions issues and typos (`sumbit_user_class`), since posting #385 I realised a big drawback of having different endpoints, which is that WTForms doesn't pass along form errors correctly anymore.

A different approach is needed, namely splitting up views into different functions rather than just different endpoints, or even rewriting the views in a class-based manner to reduce verification boilerplate and get a more declarative way of what a specific handler should check.

Thanks for the work you've done though, I appreciate you taking a stab at this complex task and pushing the conversation forward. However this appears to be the wrong approach, so I'm going to reject this. I'll be adjusting #385 accordingly.

PR #572, opened 2019-06-11
Adding UX 'enhancement' for VOSTFR Torrents

marwanpro:
I think that this improvement can improve the current disagreements between members of the French community.

As a result, this will reduce the burden on moderators in the long term.


Arylide:
This repo is only about the software, not any specific site using it.

marwanpro:
You didn't understand the commit, I think it can be useful for any deployment of Nyaa instance and for everyone. ^^

PR #575, opened 2019-06-25
Add toggle for fixed navbar

### LINKED ISSUE #274: How about making top navbar sticky while browsing the page?

cirelin:
Resolves #274 

Added a link to toggle the navbar to be static or fixed and addressed the issue where the mobile navbar would overflow in fixed mode.

This would make the fixed navbar available for those who would like to have it.

### DIFF
--- a/nyaa/static/css/main.css
+++ b/nyaa/static/css/main.css
@@ -220,9 +220,11 @@ body.dark table.torrent-list > tbody > tr.danger > td:nth-child(7) {
 }
 
 /* Allows the bootstrap selects on nav show outside the
- collapsible section of the navigation */
+ collapsible section of the navigation and removes any
+ fixed maximum height applied */
 .navbar-collapse.in {
 	overflow-y: visible;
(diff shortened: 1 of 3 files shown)


CounterPillow:
I'd rather not have this add more stuff to the footer. Maybe it's time we come up with some preferences menu accessible to people without an account.

Arylide:
Thank you for the PR, however I don't think it will go anywhere as we're not interested in a fixed navbar option at this point anyway, the issue should probably have been closed but we don't really actively go over old issues so I didn't bother. My reasoning for why is really just that nobody actively involved in the project cares enough to maintain a lot of potential options people might PR in if we accept anything, so I'd rather just say no now than deal with eventually removing a feature later. I do not see this as a game breaking feature someone absolutely needs and cannot use the site without, and I'm highly against implementing such features. I'm sorry you wasted may have wasted your time, in the future you might consider commenting on the issue asking if there is interested before working on it to avoid wasted efforts.

PR #576, opened 2019-07-01
elasticsearch 7.x compatability

qqueue:
i got it to work on my machine, including the binlog thing. Nothing major appears to have changed besides their mapping syntax and the type removal thing. Maybe there are better search features now though, I'd check the changelog at some point

https://www.elastic.co/blog/elasticsearch-7-0-0-released
https://www.elastic.co/blog/elasticsearch-6-0-0-released

### DIFF
--- a/create_es.sh
+++ b/create_es.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -e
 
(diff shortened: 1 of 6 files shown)


Arylide:
Seems to work, so will merge but would like if @TheAMM could eyeball the sync_es script.

PR #585, opened 2019-09-03
Follow Google's example for less offensive code

CounterPillow:
Through an etymological misinterpretation arising from being paid $9999999 a year to find racism where there is none, [the finest 10x engineers at Google][1] have determined that the terms "bl\*cklist" and "wh\*telist" are offensive, no doubt while sipping a latte in a gentrified neighbourhood café.

Therefore, we should replace the term "black" with a word that is universally acknowledged to have negative connotations, namely "gamer".

[1]: https://chromium.googlesource.com/chromium/src.git/+/a9d6aa6ffdfea124df722090104e0e491d6bf071%5E%21/

:^)


collinoswalt:
This change breaks functionality; nothing will be added to these lists because gamers are dead

PR #595, opened 2019-10-31
Allow filtering by multiple users

TheAMM:
⚠️ Merge after #592 ⚠️ 

```
...by repeating &u=user in the GET parameters.
No UX for this yet.
Reworks the RSS URL generator to fit with duplicated keys.
```

Sort of tested but could use a sharp look at.
```py
same_user = False
if logged_in_user:
    same_user = len(user_ids) == 1 and logged_in_user.id in user_ids
```
should replicate the previous functionality well enough.

### DIFF
--- a/config.example.py
+++ b/config.example.py
@@ -156,6 +156,13 @@ RESULTS_PER_PAGE = 75
 # How many pages we'll return at most
 MAX_PAGES = 100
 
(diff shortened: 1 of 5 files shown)


Vodes:
Is this ever going to be a thing? I'd rather not get ratelimited trying to avoid trolls ^^

PR #620, opened 2020-08-23
Support webtorrent protocol by adding webtorrent capable trackers

Banou26:
fix https://github.com/nyaadevs/nyaa/issues/621

This change add two of the recommended webtorrent trackers `https://openwebtorrent.com/` &`https://btorrent.xyz/`

The WebTorrent protocol is quite widely supported amongst popular webtorrent clients so this will enable peers to seed to browser clients that only support webtorrent peers.


ReimuHakurei:
The only trackers that will be added by default are those that are trusted by the staff. As I don't think any of the project staff had ever even heard of either of these trackers prior to this PR, they will not be added. Uploaders are free to add them to their torrents if they want to, though.

